PR #5, opened 2024-11-25
Improve macro hygiene

aecsocket:


Shatur (commented):
I agree with the idea of not relying on the prelude import, but not sure if it's the best way. Not very knowledgeable in macros,  tho :sweat_smile: 

I saw that `bevy_atmosphere` and `leafwing-input-manager` use [proc_macro_crate](https://docs.rs/proc-macro-crate/latest/proc_macro_crate), maybe do the same?

There is also [bevy_macro_utils::BevyManifest](https://docs.rs/bevy_macro_utils/latest/bevy_macro_utils/struct.BevyManifest.html).

What do you think?

aecsocket:
I've changed this to:
- `extern crate self as bevy_enhanced_input` in `lib.rs` so all the unit tests compile
- add the leading `::` to the macro path
But I think, if a user uses this crate and renames `bevy_enhanced_input` to something else in the manifest, the derive macro will break. Not sure what to do about this.

Shatur (approved):
> I don't think this needs something that heavyweight which parses the Cargo.toml.

Agree, it's a bit weird.

> I see that bevy_reflect uses BevyManifest

It works the exact same way as `proc-macro-crate`.

> But I think, if a user uses this crate and renames bevy_enhanced_input to something else in the manifest, the derive macro will break.

Yeah, this is what `proc-macro-crate` and `bevy_macro_utils` trying to solve. And this is why I decided to ask. I personally agree with the current PR state :)

Shatur:
Thanks for looking into it!

PR #6, opened 2024-11-25
Typed event values

aecsocket:


aecsocket:
Thoughts on the overall structure? Still need to write docs for the new items.

I think in general, this entire approach can be simplified. But I'm not entirely sure how.

Shatur (commented):
I like it!

Except aren't we decided to go with `output = Vec2` in Discord?
I like the ergonomic of `dim = Axis2D` and autocompletion, but I don't think it outweighs the benefits of simplicity of just `output = Vec2` :thinking:

aecsocket:
Changed to use the output type directly.

src/input_context/input_action.rs
```diff
@@ -254,27 +251,32 @@ pub enum ActionState {
...
+    /// - Use [`f32`] for single-axis actions (e.g., `Zoom`).
+    /// - For multi-axis actions, like `Move`, use [`Vec2`] or [`Vec3`].
     ///
-    /// Use [`ActionValueDim::Bool`] for button-like actions (e.g., `Jump`).
-    /// Use [`ActionValueDim::Axis1D`] for single-axis actions (e.g., `Zoom`).
-    /// For multi-axis actions, like `Move`, use [`ActionValueDim::Axis2D`] or [`ActionValueDim::Axis3D`].
-    const DIM: ActionValueDim;
+    /// The type here will determine the type of the `value` field on events
+    /// e.g. [`Fired::value`], [`Canceled::value`].
+    type Output: ActionValueOutput;
```
Shatur, line 279: Maybe just require `Into<ActionValue>`? It's already implemented for `bool`, `f32` and Vec's.

aecsocket: `ActionValueOutput` is used for multiple reasons:
- it can be sealed, so only `bool`, `f32`, `Vec2` and `Vec3` implement it, since those are the only `ActionValue` variants
- `const DIM` is needed to get the `ActionValueDim` of the type
- `convert_from(ActionValue)`  but this can be replaced with `Into<ActionValue>`

src/input_context/context_instance.rs
```diff
@@ -439,6 +439,8 @@ impl GamepadStick {
 
 #[cfg(test)]
 mod tests {
+    extern crate self as bevy_enhanced_input;
```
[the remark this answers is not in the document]
aecsocket: Leftover from merging in master :pensive:

src/input_context/events.rs
```diff
@@ -123,11 +123,8 @@ impl<A: InputAction> Copy for Ongoing<A> {}
 /// this event is triggered when the user releases the key.
 #[derive(Debug, Event)]
 pub struct Fired<A: InputAction> {
-    /// Action for which the event triggers.
-    pub marker: PhantomData<A>,
-
     /// Current action value.
-    pub value: ActionValue,
+    pub value: A::Output,
```
Shatur, line 127: There are other event structs in this file, I think we need to change them as well.

aecsocket: Didn't notice those, fixed

src/input_context/input_action.rs
Shatur: Agree, let's keep a separate trait.
But do we need to seal it? If someone wants to use another type, is there a possible footgun? :thinking: I would remove sealing and for conversion just use the `Into` impl to avoid duplication (just add the requirement for the trait).

aecsocket: I suppose someone *could* use another type, and impl `ActionValueOutput` manually. But I don't see any real reason to do that, and this would also mean that `ActionValueOutput` is part of the public API and would have to follow semver. So we couldn't modify its internals without it counting as a breaking change.

I see this trait as more of an implementation detail, so I would keep it sealed. Maybe even move the public const and fn into the `sealed` trait.

Shatur: > But I don't see any real reason to do that

Me too. But I prefer to seal types only if using them might cause a footgun. Sealed types just don't play nice with the docs. Since it's a very simple trait, I would allow user to see it, just don't add it to the prelude.

Shatur (approved):
Thanks! I just put a few tiny changes, outlined the reasoning in the commits description. Merging.

PR #7, opened 2024-11-25
BindPreset trait

aecsocket:


src/input_context/preset.rs
```diff
@@ -0,0 +1,149 @@
...
+    fn with_condition<C: InputCondition + Clone>(self, condition: C) -> PresetWithCondition<Self, C>
+    where
+        Self: Sized,
+    {
+        PresetWithCondition {
+            preset: self,
+            condition,
+        }
+    }
+}
```
Shatur, line 27: These two methods could confuse users. For example, calling `with(GamepadStick::Left.with_modifier(DeadZone::default())`, won't work as expected. The only use case for them I could think of to is applying a modifier :thinking: 

We could fix the mentioned footgun by adding native support for bind groups. So instead of having only the input-level and action-level groups, we could allow input-level groups to be nested in order to share modifiers. This will also unlock a syntax like this:

```rust
ctx.bind::<DummyAction>()
    .with((Input::mouse_motion(), GamepadString::Right).with_modifier(DeadZone::default()));
```

But I think it's outside of the scope of the PR.

Shatur (changes requested):
Looks good, left a few suggestions.

src/input_context/preset.rs
```diff
@@ -0,0 +1,149 @@
+use bevy::prelude::*;
+
+use crate::{Input, InputBind, InputCondition, InputModifier, Negate, SwizzleAxis};
+
+pub trait BindPreset {
+    fn bindings(self) -> impl Iterator<Item = InputBind>;
```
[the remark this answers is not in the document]
aecsocket: Functions should typically accept `Into`-likes, and never return `Into`-likes. Since, if `bindings` returns an `IntoIterator`, there's nothing that a caller can do with that function except for call `into_iter` on it. This step should be the implementor's responsibility.

src/input_context/preset.rs
```diff
@@ -0,0 +1,149 @@
+use bevy::prelude::*;
```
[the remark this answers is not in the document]
aecsocket: Moved to preset module

src/input_context/preset.rs
aecsocket: Agreed, I was a bit skeptical of including these fns as well, but this can always be added later.

src/input_context/preset.rs
```diff
@@ -0,0 +1,140 @@
...
+/// Maps 4 buttons as 2-dimentional input.
+///
+/// This is a convenience preset that uses [`SwizzleAxis`] and [`Negate`] to
+/// bind the buttons to X and Y axes.
+#[derive(Debug, Clone, Copy)]
+pub struct XyAxis {
+    pub up: Input,
+    pub left: Input,
+    pub down: Input,
+    pub right: Input,
```
Shatur, line 24: Small thing to consider.

In my gave I have keyboard settings defined like this:
https://github.com/projectharmonia/project_harmonia/blob/271a425fab59bcf93fabc2a6e3ee8f32940f08be/base/src/settings.rs#L119-L123

They are `Vec`s since you can assign multiple inputs to moving forward, for example.

And because of this I don't use presets in my game, I just assign them manually:
https://github.com/projectharmonia/project_harmonia/blob/271a425fab59bcf93fabc2a6e3ee8f32940f08be/base/src/game_world/player_camera.rs#L228-L243

And I wondering if your preset suggestion could address this. I think we can turn `XyAxis` into `XyAxes` and accept iterators for each direction. This way I could convert keys from settings into the preset. Other presets will be unchanged. What do you think? Is it worth it? Or maybe you have a different idea?

aecsocket: I disagree with `with` changing to `with_preset`. `with(KeyCode::W)` feels very natural - `with(GamepadStick::Left)` should feel just as natural. Same for `with(WasdKeys)`.

If I have to write `with_preset(WasdKeys)`, I'm thinking, "what the hell is a preset? I just want to `bind` `Move` to `WasdKeys`." -> `ctx.bind::<Move>().with(WasdKeys)`

Re your game code, this is a great example of DX we can improve. We should do 2 things:
- `impl BindPreset for Vec<InputBind>`
  - more generally, `for impl IntoIterator<impl Into<InputBind>>`
- `XyAxes<Up, Left, Down, Right>` where each type parameter is itself an `impl BindPreset`

Then we can turn this code:
```rs
        for &key in &settings.keyboard.camera_forward {
            move_action.with(InputBind::new(key).with_modifier(SwizzleAxis::YXZ));
        }
        for &key in &settings.keyboard.camera_left {
            move_action.with(InputBind::new(key).with_modifier(Negate::default()));
        }
        for &key in &settings.keyboard.camera_backward {
            move_action.with(
                InputBind::new(key)
                    .with_modifier(SwizzleAxis::YXZ)
                    .with_modifier(Negate::default()),
            );
        }
        for &key in &settings.keyboard.camera_right {
            move_action.with(key);
        }
```
into
```rs
move_action.with(XyAxes::new(
    &settings.keyboard.camera_forward,
    &settings.keyboard.camera_left,
    &settings.keyboard.camera_backward,
    &settings.keyboard.camera_right,
});
```

Also, there should be an example which uses code similar to your game's control settings. It demonstrates a great pattern which would be commonly used by others.

Shatur: Hm... Not sure about `with` vs `with_preset... I think that the distinction between inputs and presets are quite logical. It just creates another level of structs indirection that we pass to this function, but without much practical benefit. Let me ask one more person opinion about it.

Agree with the `XyAxes` suggestion!

Shatur: > north/south/east/west? I

Agree!

src/input_context/bind.rs
```diff
@@ -0,0 +1,164 @@
...
+                BindConfigSet::from_iter([
+                    $(self.$i.into_configs()),*
+                ])
+                .into_configs()
+            }
+        }
+    };
+}
+
+impl_tuple!(A 0, B 1);
```
Shatur, line 150: Bevy already provides a helper for it: https://docs.rs/bevy/latest/bevy/utils/macro.all_tuples.html
Usage example from replicon: https://github.com/projectharmonia/bevy_replicon/blob/df44894205625e6e3dd8f627d6651545d12d3030/src/core/replication/replication_rules.rs#L329

examples/from_settings.rs
```diff
@@ -0,0 +1,108 @@
+//! Creates a context from a settings struct.
```
Shatur, line 1: I think this should be just a doc example for `XyAxes`. No the whole main with systems, just the settings struct and preset creation from it to showcase it.

src/input_context/bind.rs
aecsocket: I tried using this, but it doesn't work here because we need access to the index of the field we're using as well, i.e. in `(T0, T1)` we need `0, 1`. So I had to implement this manually.

examples/from_settings.rs
aecsocket: Added a doc example to `Cardinal` and deleted this file

src/input_context/bind.rs
Shatur: Ah, I see!

src/input_context/preset.rs
```diff
@@ -0,0 +1,173 @@
...
+/// use bevy::prelude::*;
+/// use bevy_enhanced_input::prelude::*;
+///
+/// #[derive(Debug, Resource)]
+/// struct ControlSettings {
+///     forward: Vec<Input>,
+///     right: Vec<Input>,
+///     backward: Vec<Input>,
+///     left: Vec<Input>,
+/// }
```
Shatur, line 27: Let's make the example more common with a clarification:
```suggestion
// In games different input sources usually have different bindings.
// But also possible to `Input` to consolidate different input sources in one setting.
/// #[derive(Debug, Resource)]
/// struct KeyboardSettings {
///     forward: Vec<KeyCode>,
///     right: Vec<KeyCode>,
///     backward: Vec<KeyCode>,
///     left: Vec<KeyCode>,
/// }
```

src/input_context/bind.rs
```diff
@@ -0,0 +1,172 @@
...
+}
+
+impl Extend<BindConfigs> for BindConfigSet {
+    fn extend<T: IntoIterator<Item = BindConfigs>>(&mut self, iter: T) {
+        self.binds.extend(iter)
+    }
+}
+
+#[derive(Debug)]
+pub struct BindConfig {
```
Shatur, line 31: Since we removing `InputBind`, maybe re-use it name here? :thinking: I.e. InputBindSet, InputBinds, etc.

Just not a fan of the word "config" in this case.

src/input_context/bind.rs
```diff
@@ -0,0 +1,172 @@
+use crate::prelude::*;
+
+#[derive(Debug, Default)]
+pub struct BindConfigInfo {
+    pub modifiers: Vec<Box<dyn InputModifier>>,
+    pub conditions: Vec<Box<dyn InputCondition>>,
+}
```
Shatur, line 7: Maybe inline this two? Yes, it will require adding two more methods for getting it, but calling `self.conditions()` is nicer then `self.info().conditions`.

src/input_context/preset.rs
```diff
@@ -0,0 +1,173 @@
...
+///
+/// #[derive(Debug, Component)]
+/// struct PlayerInputContext;
```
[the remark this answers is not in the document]
aecsocket: Changed to `Player`

src/input_context/preset.rs
aecsocket: Changed and added comments

src/input_context/bind.rs
aecsocket: agreed, renamed

src/input_context/bind.rs
aecsocket: agreed, inlined into the structs themselves

Shatur:
I split the logic from this PR into multiple chunks and merged them into master.

Not done improving the ergonomic yet, I planning to add tuple syntax and other improvements tomorrow.

PR #9, opened 2024-11-26
Add `InputContext` derive

aecsocket:
```rs
#[derive(Component, InputContext)]
#[input_context(ctor = new_dummy_context)]
struct DummyContext;

fn new_dummy_context(In(_): In<Entity>) -> ContextInstance {
    let mut ctx = ContextInstance::default();
// ...
}
```

### DIFF
--- a/examples/all_conditions.rs
+++ b/examples/all_conditions.rs
@@ -34,10 +34,10 @@ impl GamePlugin {
 }
 
 #[derive(Component, InputContext)]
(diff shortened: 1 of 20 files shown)


Shatur:
Discussed in Discord, the change sounds cool, but doesn't really improve the ergonomic...
Keeping the current approach, it's similar to `FromWorld`.

PR #12, opened 2024-12-01
Fix typos and add `typos` to CI

aecsocket:
https://github.com/crate-ci/typos


Shatur (approved):
Thank you a lot! Will add to all my other repos :sweat_smile:

Shatur:
Ah, looks like I rushed a bit with the `Negate` PR, doc tests fail. But this one can be merged as is.

Issue #24, opened 2025-01-30
Add keybinding menu example w/ serialization


alice-i-cecile:
Must use bevy_ui

Shatur:
Agree! I remember implementing it with egui for LWIM and submitting it, but it was hard to maintain 😅

Shatur:
I thinking about porting settings menu from my game, it's pure bevy_ui. Just remove tabs and replace Ok and Cancel with Apply.

![Image](https://github.com/user-attachments/assets/e6ba59ef-eb85-4330-9c2b-a918f12fca1c)
![Image](https://github.com/user-attachments/assets/b650f916-4113-4527-85eb-3e637902caad)
![Image](https://github.com/user-attachments/assets/4323e53d-eaa8-4fe6-9409-d1aa155e4d92)

alice-i-cecile:
That would be perfect :D

PR #34, opened 2025-01-31
Derive `Reflect` for `Input` and `ModKeys`

Shatur:
For `ModKeys` it requires macro separation as described in
https://docs.rs/bitflags/latest/bitflags/#custom-derives

Need this change to implement #24. In my game I planning to have dedicated tabs for gamepad and keyboard inputs. But for the example simplicity I want to have a single menu where user can bind any input source.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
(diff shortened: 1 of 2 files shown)


alice-i-cecile (approved):
I think it probably makes sense to feature flag all of the reflection, but this is fine for this PR.

Shatur:
Since we currently depend on `bevy`, `bevy_ecs` pulls in `bevy_reflect`.
However, when/if we start preparing for upstreaming, we'll definitely need to feature-flag it!

PR #37, opened 2025-02-02
Add `InputEvent`

Shatur:
Can be used to display appropriate button icons or bind any input sources to actions in settings. In my game I only bind keyboard inputs (and planning to bind gamepad inputs in a separate settings tab), but for the example I want to showcase any `Input` serialization.
Emitted by separate `InputEventPlugin`.

To organize it nicely, I moved current `EnhancedInputPlugin` under `input_context` module and renamed it into `InputContextPlugin` since it's responsible only for updating contexts.
And added `EnhancedInputPlugins` (plural) group that contains both `InputEventPlugin` and `InputContextPlugin`.

Both plugins are optional and can function without each other.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Derive `Reflect` for `Input` and `ModKeys`.
+- Derive `PartialEq` for `Input`.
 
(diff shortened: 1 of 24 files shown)


alice-i-cecile (approved):

Shatur:
Forgot to cherry-pick e09a672 from my local working branch :sweat_smile:
Should be good now.

If you have any suggestions about anything, including naming and code organization - feel free tell, I always open to other people opinions :)

Issue #22, opened 2025-01-30
Add a way to detect what input modality is used


alice-i-cecile:
Keyboard + Mouse, gamepad etc. Great for prompts. Store in a resource.

Shatur:
Reopened due to #39.

alice-i-cecile:
Yeah, just providing a simple resource for users to read would be ideal. TBH this can / should probably just live in bevy_input eventually.

Shatur:
Maybe it should be directly submitted it to `bevy_input`?.. It doesn't need any API from this crate.

alice-i-cecile:
Agreed; let's migrate this issue.

PR #38, opened 2025-02-02
Several API changes: rename InputActionState back to EventReader and revert InputEvent

Shatur:
I am working on the keybindings menu example (and reworking the keybindings in my own game as well), and this PR contains several improvements that I need in order to submit it:
- I realized that `InputEvent` is not a good idea, so I reverted it. It's not very useful for settings because you don't want to listen for any input event. It doesn't make much sense to listen for keyboard inputs, gamepad axes, or mouse movements, as they all have different arity. Yes, it also solves #22, but I think it's better to provide a resource, as was originally suggested. On top of that, having another input event will be confusing for users.
- Reverted the ignoring of `clippy::too_many_arguments` (no longer needed after the `InputEvent` revert).
- Renamed `InputActionState` back to `InputReader`. I experimented with it and think that the original name was better after all. However, I kept its new location under the `input_context` module.
- Implemented `Display` for `Input`. It's very convenient to have.
- Reordered `ModKeys` constants to simplify the `Display` impl.
- Added `must_use` for `ModKeys::pressed` to avoid potential mistakes.

Each commit is a separate change.


alice-i-cecile (approved):
I think this is a reasonable set of changes. Particularly agree on removing InputEvent.

Shatur:
Thanks!
I usually try to submit separate PRs, but I was working on the example and didn't expect you to take a look into it at Sunday :)

PR #40, opened 2025-02-03
Add keybindings menu example

Shatur:
Short showcase video:

https://github.com/user-attachments/assets/5b63f3b4-7fb1-4dc3-8566-e24dbc67cc4a

### DIFF
--- a/examples/ui_priority.rs
+++ b/examples/ui_priority.rs
@@ -1,5 +1,5 @@
 //! Inputs consumed by UI and not propagated to actions.
-//! In order to run this example pass `--features egui_priority,bevy_egui/render,bevy_egui/default_fonts,bevy/default_font` to cargo.
+//! In order to run this example pass `--features egui_priority,bevy_egui/render,bevy_egui/default_fonts` to cargo.
 
 mod player_box;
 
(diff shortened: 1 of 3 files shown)


examples/ui_priority.rs
```diff
@@ -1,5 +1,5 @@
 //! Inputs consumed by UI and not propagated to actions.
-//! In order to run this example pass `--features egui_priority,bevy_egui/render,bevy_egui/default_fonts,bevy/default_font` to cargo.
+//! In order to run this example pass `--features egui_priority,bevy_egui/render,bevy_egui/default_fonts` to cargo.
```
alice-i-cecile, line 2: FYI I think you can specify this in the example metadata. Check what the Bevy cargo.toml is doing!

alice-i-cecile (approved):
Excellent :D I'll do another pass on the examples later, but I'm really pleased to have something to work off of.

examples/ui_priority.rs
Shatur: Yep, I configured it:
https://github.com/projectharmonia/bevy_enhanced_input/blob/9727118af46a6daed3ee11f97ea0e197e8d3745e/Cargo.toml#L42
Cargo just don't tell users which command to run, so I decided to clarify :)

PR #42, opened 2025-02-08
Fix some comments that should be doc

Zoomulator:
Changed some field comments that were left as regular comments, as they appear to be intended as public documentation.


Shatur (approved):
Thanks a lot!

PR #43, opened 2025-02-09
Remove `InputContext` trait

Shatur:
The trait holds only a context creation function and priority. Both of these things can be set at registration time.

Originally, I planned to turn the context creation function into a system. However, I needed to pass an entity into it and return a context from it. Using `In` and chaining it with another system to retrieve the return value didn’t feel very clean, so I decided to make it an observer instead.

Observers can hold an entity and allow mutating the trigger. So, I introduced `Bindings<C>`, which is triggered on component insertion and dereferences into an `InputContext`. Users can mutate the context in observers, and the value is then moved into the registry and evaluated as usual.

See the changelog for details about the renames I had to do to keep things organized.

The diffs for `input_context` look weird because I renamed `context_instance` into `input_context` and old `input_context` into `registry`. The same with `tests/input_context.rs` -> `tests/registry.rs`. Hence the large diff line count.

These changes also make the registry logic much simpler. It’s just a `Vec` of instances associated with entities and their component `TypeId`s.

The priority is set on the `Bindings` trigger. I decided to make it like in Unreal - you can optionally assign the priority on input context creation based on any logic.

@alice-i-cecile don't afraid to criticize the API, I want to hear you opinion on the changes :slightly_smiling_face:


alice-i-cecile (approved):

alice-i-cecile:
I'm generally very pleased with this: I think it's a critical step towards simplifying the end user API. Let's merge this and then refine in follow-up: the large diff really bogs down Github for me.

PR #45, opened 2025-02-11
Fix `Cardinal` directions

### LINKED ISSUE #44: `Cardinal` is mirrored incorrectly
`Cardinal` negates the east input and doesn't negate the west input. Additionally, gamepads bind east to d-pad left and west to d-pad right. Same for wasd. This is backwards as far as I know. I'd expect east to be right and west to be left.

jnhyatt:
Fixes #44

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Derive `Reflect` for `Input` and `ModKeys`.
 - Derive `PartialEq` for `Input`.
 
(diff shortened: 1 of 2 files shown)


alice-i-cecile (approved):

Shatur:
Thanks a lot!
The names were just swapped, which is why the examples and tests worked as expected :sweat_smile:
I adjusted the ordering in the PR to make it more noticeable in the diffs.

I'm planning to draft a new release soon. Just want to include #43 in it, so I'm just waiting for the review. Let me know if you have any thoughts on it!

PR #52, opened 2025-03-16
Prepare for Possible `no_std` Support

bushrat011899:
# Objective

With Bevy 0.16, there is the possibility of many crates gaining `no_std` support. Further, `no_std` support is desirable for crates that will be upstreamed.

## Solution

- Added lints to reduce reliance on `std`.
- Switched to `core::prelude` instead of `std::prelude` using `#![no_std]`.
- Added `std` and `alloc` as external crates.
- Disabled default features on `serde` and re-enabled features used directly by this crate.

---

## Notes

Once Bevy 0.16 releases, `no_std` support should be as simple as:
- Adding an `std` feature and including it in `default`
- Gating the `extern crate std;` statement
- Gating the `ui_priority` and `egui_priority` features on `std`
- Potentially adding other support features such as `libm`, `critical-section`, and `web` for ease of use.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,6 +48,9 @@ required-features = [
 
 [lints.clippy]
 type_complexity = "allow"
+alloc_instead_of_core = "warn"
+std_instead_of_alloc = "warn"
+std_instead_of_core = "warn"
 
 [workspace]
 members = ["macros"]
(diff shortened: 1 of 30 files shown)


Shatur (approved):
That was quick :sweat_smile:
Thanks!

No sure why cargo check started to fail, but I don't think it's related to the changes.

bushrat011899:
I've been getting fast at this! Plus this crate was already very close to `no_std` compatible anyway

PR #53, opened 2025-03-17
Capture gamepad axes as `Axis1D`

Shatur:
Because triggers are also buttons and sometimes have analog value.


alice-i-cecile (approved):
I'm still not thrilled with doing this sort of thing. As far as I can tell, this is *only* used for triggers: 90% of the time you want to handle gamepad buttons as purely binary data. That said, I'm not sure if there's a better design here.

Shatur:
Same :slightly_smiling_face:
But it shouldn't hurt the ergonomics because for boolean actions it will be automatically converted. And user can control the actuation via conditions.

Issue #25, opened 2025-01-30
Rewrite README to make core concepts clearer and seem simpler


alice-i-cecile:

Shatur:
That would be great! I'm also not a native speaker, but I tried my best. My native language is quite different :slight_smile:

PR #60, opened 2025-03-29
Component-based API

### LINKED ISSUE #59: context_switch example did not work as expected
Hi, 
i just cloned the repo (master branch) and run the `context_switch` example (`cargo r --example context_switch`) and hit the Enter key twice. The application window terminates with the error message below.  It seems, that the `OnFoot` wasn't removed on the `enter_car` event. The context_layering example has the same problem. Unfortunately I don't have enough experience with Bevy to analyze the error myself.

`
thread 'main' panicked at /home/ld/Repository/Rust/bevy/bevy_enhanced_input/src/registry.rs:114:5:
context for `context_switch::OnFoot` shouldn't already exist on `21v1`
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

2025-03-27T12:02:02.149688Z  WARN bevy_ecs::world::command_queue: CommandQueue has un-applied commands being dropped. Did you forget to call SystemState::apply?
Encountered a panic in system `bevy_app::main_schedule::Main::run_main`!
`

- Bevy 15.3
- openSUSE Tumbleweed x86_64 - wayland
- nightly-x86_64-unknown-linux-gnu

Shatur:
## Issues with the current design

- While the trigger-based (push-style) API is awesome, we have an unergonomic pull-style API. Users need to access the registry, then obtain the context, and only then access data for actions. Users are often confused by this and assume that the trigger-based API is the only API.
- Defining bindings on a trigger is ergonomic, but due to the lack of ordering for observers in Bevy, it sometimes leads to unexpected behavior like #59.
- We need to associate a schedule for context in the future, and there is no convenient way to do this with the current API.

## Solution

Instead of using a registry with contexts, we now store data directly on entities. Since there could be multiple contexts, I also added a marker. But `InputContext<Marker>` is quite a long name. I remember @alice-i-cecile disliked the `InputContext` name, but we couldn't find an alternative. In this PR, I renamed `InputContext` to just `Actions` - I think this name fits now!

However, we still need the registry for two reasons:
- We need to iterate over actions in a defined order.
- We need to work with actions in a type-erased way.
But now it's an implementation detail; we don't expose our registry to users anymore.

I also mentioned that we need a way to nicely configure actions. Since we now have markers, I moved the configuration to the newly added `ActionsMarker` trait and provided a derive macro. Right now it's only `priority`, but I'm planning to add `schedule` in the future.

Since actions are now components, I reworked how the `Bindings` trigger works. Users can now just query the actions and mutate them directly. This is more idiomatic and fixes #59. It's a bit less ergonomic, but I think this could be improved on Bevy's side since it's a very common pattern.

## API showcase

Before:

```rust
app.add_input_context::<Player>().add_observer(player_binding);

fn player_binding(mut trigger: Trigger<Binding<Player>>) {
    trigger.bind::<Jump>().to(GamepadButton::South);
}

#[derive(Component)]
struct Player;

#[derive(Debug, InputAction)]
#[input_action(output = bool)]
struct Jump;
```

After:

```rust
app.add_actions_marker::<Player>().add_observer(player_binding);

fn player_binding(
    trigger: Trigger<Binding<Player>>,
    mut players: Query<&mut Actions<Player>>,
) {
    let mut actions = players.get_mut(trigger.entity()).unwrap();
    actions.bind::<Jump>().to(GamepadButton::South);
}

#[derive(ActionsMarker)]
struct Player;

#[derive(Debug, InputAction)]
#[input_action(output = bool)]
struct Jump;
```

Previously, to access values in a system, users needed to read the `InputContextRegistry` resource:

```rust
fn observer(trigger: Trigger<Attacked>, registry: Res<InputContextRegistry>) {
    let ctx = registry.context::<Player>(trigger.entity());
    let value = ctx.action::<Dodge>().value();
    // ..
}
```

Now it's just a regular query:

```rust
fn system(trigger: Trigger<Attacked>, players: Query<&mut Actions<Player>>) {
    let actions = players.get_mut(trigger.entity()).unwrap();
    let value = actions.action::<Dodge>().value();
    // ..
}
```

## Follow-up work

To make the PR actually reviewable for @alice-i-cecile, I tried to avoid renames - except for essentials like `InputContext` -> `Actions` to better present the API.

I'm planning to overhaul names and reorganize things in follow-ups without functional changes. For example, `actions.action::<Dodge>` is repetitive; I’d prefer `actions.get::<Dodge>` and similar improvements.

The same applies to docs. We have issue #25 for a planned rework, but before @alice-i-cecile assists, I wanted to iterate on it myself first. I recently did the same thing for `bevy_replicon` and quite happy with the [end result](https://docs.rs/bevy_replicon/latest/bevy_replicon) :slightly_smiling_face:

## Final notes

Don't be scared by the huge diff. Most of the changes are from "context" to "actions" and related docs/examples changes to the new API. I didn't touch the internal logic, only how the registry operates on actions, which is in the `action_instances` module, and how systems read actions, which is in `lib.rs`.

I would appreciate feedback on the new API and especially on the new naming.


jnhyatt:
One bit of feedback from migrating my project to use this:
```rs
#[derive(ActionsMarker)]
#[require(FlycamSpeed)] // ERROR
pub struct Flycam;
```
I can't use the `Component` derive ecosystem anymore. I'd suggest not auto-deriving `Component` in the `ActionsMarker` derive, and instead just rely on the supertrait bound telling the user they need to also implement `Component`. In the end, I think it looks fine:
```rs
#[derive(Component, ActionsMarker)]
#[require(FlycamSpeed)]
pub struct Flycam;
```
An alternative would be to just try and copy all the `Component` derive attributes into the `ActionsMarker` derive, but then you have to try to keep it in sync with each Bevy version and remember to check for `Component` derive changes every new release. Sounds like a pain.

Shatur:
@jnhyatt it no longer needs to be a component. It's just a marker for `Actions<T>`, which is a component. You insert `Actions<T>`, not `T` itself :slightly_smiling_face:

alice-i-cecile (approved):
Excellent! This solves two of my biggest problems with the crates (naming, bad pull-style API): I'm really pleased that you could get this working.

PR #65, opened 2025-04-04
Reorganize crate

Shatur:
The refactoring I promised. We had a lot of iterations on the API, so it's time to revisit the naming.  
This PR contains only renaming and reorganization, with no logical changes.

Since looking at diffs would be messy, I put each change into a separate commit with a description to simplify the review. This will also make it easy to revert if we disagree on something.  

Let me know what you think about all these changes :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Input contexts no longer associated with a component. Users need to manually insert `Actions`.
 - Replace `InputContextAppExt::add_input_context` with `ActionsMarkerAppExt::add_actions_marker` that accepts markers instead of components.
 - `Binding` trigger no longer stores bindings. Just get `Actions` using `Query` and mutate it directly.
+- Rename `ActionBind` into `ActionBinding` and move into `action_binding` module.
+- Rename `Actions::action_bind` and `Actions::get_action_bind` into `Actions::binding` and `Actions::get_binding`.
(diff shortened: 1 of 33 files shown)


alice-i-cecile:
> [Use crate:: instead of super:: for top-level modules](https://github.com/projectharmonia/bevy_enhanced_input/pull/65/commits/7c23ab20f385dfa7fa37190eb108564f36389725) 

Great! More robust.

> [Use pub(crate) for top-level modules](https://github.com/projectharmonia/bevy_enhanced_input/pull/65/commits/9fec494217e6762a8beda5b1e64cea84a38ea1bf) 

Sure :)

> [Rename InputBind into InputBinding](https://github.com/projectharmonia/bevy_enhanced_input/pull/65/commits/fa055012b11b9300d4359c4ecc7e074bee1b9f50) 

Definitely more natural of a name.

> [Rename *Set traits into *Into](https://github.com/projectharmonia/bevy_enhanced_input/pull/65/commits/04fcc0d8a314df0ee03928c235a23005fd1dc811) 

Much clearer.

> [Rename InputBindModCond into BindingBuilder](https://github.com/projectharmonia/bevy_enhanced_input/pull/65/commits/adc48f8c4c3f6c5aeedabd945423f90b895f88fe) 
adc48f8

Great; abbreviations are frustrating.

> [Rename ActionBind into ActionBinding and move into a dedicated mo…](https://github.com/projectharmonia/bevy_enhanced_input/pull/65/commits/f977f1f104e20fe37509b263949b10da5c1527d9)

Same as above: binding > bind.

alice-i-cecile (approved):

PR #70, opened 2025-04-05
Multiple `ConditionKind::Blocker` fix for `TriggerTracker`

### LINKED ISSUE #69: `TriggerTracker` doesn't take into account multiple `ConditionKind::Blocker`
`TriggerTracker.apply_conditions` has the following code for `ConditionKind::Blocker`:
https://github.com/projectharmonia/bevy_enhanced_input/blob/d3f3d7c662328fecc9a48357143dfb2ba137461e/src/trigger_tracker.rs#L81-L88

If the action has multiple blocker conditions then only the last one will be considered. Should this be changed to use BitOrAssign like the other `ConditionKind` use?

Tsudico:
This should fix #69

### DIFF
--- a/src/trigger_tracker.rs
+++ b/src/trigger_tracker.rs
@@ -81,9 +81,9 @@ impl TriggerTracker {
                 ConditionKind::Blocker { events_only } => {
                     let blocked = state == ActionState::None;
                     if events_only {
-                        self.events_blocked = blocked;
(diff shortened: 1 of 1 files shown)


Shatur:
Nice catch, thanks a lot!

Shatur (approved):

alice-i-cecile (approved):

PR #71, opened 2025-04-07
Rename `ActionsMarker` into `InputContext`

Shatur:
While working on a major documentation overhaul, I noticed that `ActionsMarker` is not a good name for a trait.
In my opinion, `InputContext` fits better. Users still mostly work with `Actions`. `InputContext` is just a trait and derive name. I also like that it’s similar to the `InputAction` derive and matches Unreal terminology.

This choice might look clearer once I finish the documentation. So even if we merge this now and come up with a better term after the docs rework - I'm fully open to rename it again :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,9 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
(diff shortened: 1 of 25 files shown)


alice-i-cecile (commented):
So much clearer! Thanks for taking my complaints about clarity seriously.

alice-i-cecile (approved):

PR #73, opened 2025-04-07
Documentation overhaul

Shatur:
- Add a quick start guide that walks users through the API.
- Refine documentation on all types and ensure that it attached to a proper type after countless refactoring.
- Minor logging adjustments to use consistent terminology.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,12 +10,11 @@ Dynamic and contextual input mappings for Bevy, inspired by [Unreal Engine Enhan
 ## Features
 
 * Map inputs from various sources (keyboard, gamepad, etc.) to gameplay actions like `Jump`, `Move`, or `Attack`.
-* Assign actions to different contexts like `OnFoot` or `InCar`, which are regular components.
-* Activate or deactivate contexts by simply adding or removing components.
-* Control how actions accumulate input from sources and consume it.
+* Assign actions to different contexts like `OnFoot` or `InCar`, controlled by `Actions<C>` components.
 * Layer multiple contexts on a single entity, controlled by priority.
 * Apply modifiers to inputs, such as dead zones, inversion, scaling, etc., or create custom modifiers by implementing a trait.
 * Assign conditions for how and when an action is triggered, like "hold", "tap", "chord", etc. You can also create custom conditions by implementing a trait.
+* Control how actions accumulate input from sources and consume it.
 * React on actions with observers.
 
 ## Getting Started
--- a/src/action_binding.rs
+++ b/src/action_binding.rs
@@ -18,10 +18,24 @@ use crate::{
     trigger_tracker::TriggerTracker,
 };
 
-/// Bindings of [`InputAction`] for [`Actions`](crate::actions::Actions).
+/// Bindings assoticated with an [`InputAction`] marker.
 ///
-/// These bindings are stored separately from [`ActionMap`] to allow a currently
-/// evaluating action to access the state of other actions.
+/// Stored inside [`Actions`](crate::actions::Actions).
+///
+/// Bindings are stored separately from [`ActionMap`] to allow reading other actions' data during evaluation.
+/// The current action's data updates at evaluation's end.
+///
+/// Action bindings evaluation follows these steps:
+///
+/// 1. Iterate over each [`ActionValue`] from the associated [`Input`]s:
+///    1.1. Apply input-level [`InputModifier`]s.
+///    1.2. Evaluate input-level [`InputCondition`]s, combining their results based on their [`InputCondition::kind`].
+/// 2. Select all [`ActionValue`]s with the most significant [`ActionState`] and combine based on [`InputAction::ACCUMULATION`].
+///    Combined value be converted into [`ActionOutput::DIM`] using [`ActionValue::convert`].
+/// 3. Apply action level [`InputModifier`]s.
+/// 4. Evaluate action level [`InputCondition`]s, combining their results according to [`InputCondition::kind`].
+/// 5. Set the final [`ActionState`] based on the results.
+///    Final value be converted into [`InputAction::Output`] using [`ActionValue::convert`].
 pub struct ActionBinding {
     type_id: TypeId,
     action_name: &'static str,
@@ -158,9 +172,6 @@ impl ActionBinding {
 
     /// Adds input mapping.
     ///
-    /// The action can be triggered by any input mapping. If multiple input mappings
-    /// return [`ActionState`], the behavior is determined by [`InputAction::ACCUMULATION`].
-    ///
     /// Thanks to traits, this function can be called with multiple types:
     ///
     /// 1. Raw input types.
@@ -168,6 +179,8 @@ impl ActionBinding {
     /// 3. [`InputBinding`] which wraps [`Input`] and can store input modifiers or conditions.
     /// 4. [`IntoBindings`] which wraps [`InputBinding`] and can store multiple [`InputBinding`]s.
     ///    Also implemented on tuples, so you can pass multiple inputs to a single call.
+    ///
+    /// All assigned inputs will be treated as "any of".
     ///
     /// # Examples
     ///
--- a/src/actions.rs
+++ b/src/actions.rs
@@ -15,38 +15,12 @@ use crate::{
     input_reader::{InputReader, ResetInput},
 };
 
-/// Instance for [`InputContext`].
+/// Component that stores a actions with their bindings for specific [`InputContext`].
 ///
-/// Stores [`InputAction`]s and evaluates their [`ActionState`] in the order they are bound.
+/// Bindings represented by [`ActionBinding`] and can be added to specific action using [`Self::bind`].
+/// Data for each bound action is stored inside [`ActionMap`].
 ///
-/// Each action can have multiple associated [`Input`]s, any of which can trigger the action.
-///
-/// Additionally, you can assign [`InputModifier`]s and [`InputCondition`]s at both the action
-/// and input levels.
-///
-/// You can define bindings before the insertion, but it's recommended to create an observer
-/// for [`Binding`](crate::action_instances::Binding). To setup bindings, register an observer
-/// an obtain this component. This way you can conveniently reload bindings when you settings
-/// change using [`RebuildBindings`](crate::action_instances::RebuildBindings).
-///
-/// Until the component is exists on the entity, actions will be evaluated and trigger [`events`](super::events).
-///
-/// Action evaluation follows these steps:
-///
-/// 1. Iterate over each [`ActionValue`] from the associated [`Input`]s:
-///    1.1. Apply input-level [`InputModifier`]s.
-///    1.2. Evaluate input-level [`InputCondition`]s, combining their results based on their [`InputCondition::kind`].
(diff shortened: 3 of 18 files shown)


src/lib.rs
```diff
@@ -1,38 +1,286 @@
...
+#[derive(InputContext)]
+struct OnFoot;
+```
+
+No required parameters needed. See the [`InputContext`] documentation for all parameters.
+
+## Binding actions
+
+While input contexts are defined statically at compile time, mappings must be assigned at runtime. They're stored inside the
+[`Actions<C>`] component, where `C` is an input context. Contexts becomes becomes active only when its component exists on an
```
Tsudico, line 66: Just a grammar issue, becomes is repeated here.

src/lib.rs
```diff
@@ -1,38 +1,286 @@
...
+which applies to all inputs. For details about how multiple modifiers are merged together, see the [`ActionBinding`] documentation.
+Both methods also support tuple syntax for assigning multiple modifiers at once.
+
+```
+# use bevy::prelude::*;
+# use bevy_enhanced_input::prelude::*;
+# let mut actions = Actions::<OnFoot>::default();
+actions.bind::<Move>().to((
+    // Keyboard keys captured as `bool`, but the output of `Move` is defined as `Vec2`,
+    // so you need to assign keys to axes using swizzle and negation.
```
Tsudico, line 106: I know swizzle is understood in graphics programming, I'm not sure how widely known it is elsewhere. I wonder if it may be more clear what is being done here if it was "using swizzle to reorder them and negation"

src/input_action.rs
```diff
@@ -55,24 +28,26 @@ use crate::action_value::{ActionValue, ActionValueDim};
...
-    /// [`Canceled::value`](super::events::Canceled::value).
+    /// This type will also be used for `value` field on events
+    /// e.g. [`Fired::value`](crate::events::Fired::value),
+    /// [`Canceled::value`](crate::events::Canceled::value).
     type Output: ActionOutput;
 
     /// Specifies whether this action should swallow any [`Input`](crate::input::Input)s
     /// bound to it or allow them to pass through to affect other actions.
     ///
     /// Inputs are consumed only if the action state is not equal to
```
Tsudico, line 48:
```suggestion
    /// Inputs are consumed when the action state is not equal to
```
I think when might work better here because it is expected for the action state to change in order for the consuming input behavior to occur or not. It is the behavior when it happens that is being modified.

Is this outside the bounds of the review?

src/lib.rs
```diff
@@ -1,38 +1,286 @@
...
+and the output type will match the action’s [`InputAction::Output`]. Events also store other information, such as timings. See the
+documentation on specific event type for more information.
+
+```
+# use bevy::prelude::*;
+# use bevy_enhanced_input::prelude::*;
+# let mut app = App::new();
+app.add_observer(apply_movement);
+
+/// Apply movemenet when `Move` action considered fired.
```
Tsudico, line 246:
```suggestion
/// Apply movement when `Move` action considered fired.
```

Tsudico (changes requested):
Here is my first pass at a review.

Shatur (commented):
Thanks a ton! Agree with everything, applied all suggestions.

README.md
```diff
@@ -10,12 +10,11 @@ Dynamic and contextual input mappings for Bevy, inspired by [Unreal Engine Enhan
...
 * Map inputs from various sources (keyboard, gamepad, etc.) to gameplay actions like `Jump`, `Move`, or `Attack`.
-* Assign actions to different contexts like `OnFoot` or `InCar`, which are regular components.
-* Activate or deactivate contexts by simply adding or removing components.
-* Control how actions accumulate input from sources and consume it.
+* Assign actions to different contexts like `OnFoot` or `InCar`, controlled by `Actions<C>` components.
 * Layer multiple contexts on a single entity, controlled by priority.
 * Apply modifiers to inputs, such as dead zones, inversion, scaling, etc., or create custom modifiers by implementing a trait.
 * Assign conditions for how and when an action is triggered, like "hold", "tap", "chord", etc. You can also create custom conditions by implementing a trait.
+* Control how actions accumulate input from sources and consume it.
 * React on actions with observers.
```
alice-i-cecile, line 18:
```suggestion
* React to actions with observers.
```

src/action_binding.rs
```diff
@@ -18,10 +18,24 @@ use crate::{
...
 
-/// Bindings of [`InputAction`] for [`Actions`](crate::actions::Actions).
+/// Bindings associated with an [`InputAction`] marker.
 ///
-/// These bindings are stored separately from [`ActionMap`] to allow a currently
-/// evaluating action to access the state of other actions.
+/// Stored inside [`Actions`](crate::actions::Actions).
+///
+/// Bindings are stored separately from [`ActionMap`] to allow reading other actions' data during evaluation.
+/// The current action's data updates at evaluation's end.
```
alice-i-cecile, line 26: When does evaluation end? Where do I read more about this?

src/action_binding.rs
```diff
@@ -18,10 +18,24 @@ use crate::{
...
+///
+/// Action bindings evaluation follows these steps:
+///
+/// 1. Iterate over each [`ActionValue`] from the associated [`Input`]s:
+///    1.1. Apply input-level [`InputModifier`]s.
+///    1.2. Evaluate input-level [`InputCondition`]s, combining their results based on their [`InputCondition::kind`].
+/// 2. Select all [`ActionValue`]s with the most significant [`ActionState`] and combine based on [`InputAction::ACCUMULATION`].
+///    Combined value be converted into [`ActionOutput::DIM`] using [`ActionValue::convert`].
```
alice-i-cecile, line 34: When first encountering it I have no idea what `ActionOutput::DIM`  might be.

src/action_binding.rs
```diff
@@ -18,10 +18,24 @@ use crate::{
...
+/// 3. Apply action level [`InputModifier`]s.
+/// 4. Evaluate action level [`InputCondition`]s, combining their results according to [`InputCondition::kind`].
+/// 5. Set the final [`ActionState`] based on the results.
+///    Final value be converted into [`InputAction::Output`] using [`ActionValue::convert`].
```
alice-i-cecile, line 38: Same "be converted" complaint here :)

src/action_binding.rs
```diff
@@ -112,7 +126,7 @@ impl ActionBinding {
     /// Adds action-level conditions.
     ///
     /// For input-level conditions see
-    /// [`BindingBuilder::with_conditions`](super::input_binding::BindingBuilder::with_conditions).
+    /// [`BindingBuilder::with_conditions`](crate::input_binding::BindingBuilder::with_conditions).
```
alice-i-cecile, line 129: We should have dead doc links in CI.

src/action_binding.rs
```diff
@@ -169,6 +180,8 @@ impl ActionBinding {
     /// 4. [`IntoBindings`] which wraps [`InputBinding`] and can store multiple [`InputBinding`]s.
     ///    Also implemented on tuples, so you can pass multiple inputs to a single call.
     ///
+    /// All assigned inputs will be treated as "any of".
```
alice-i-cecile, line 183: This needs a bit more explanation to follow-up.

src/actions.rs
```diff
@@ -15,38 +15,12 @@ use crate::{
...
-/// 3. Apply action level [`InputModifier`]s.
-/// 4. Evaluate action level [`InputCondition`]s, combining their results according to [`InputCondition::kind`].
-/// 5. Set the final [`ActionState`] based on the results.
-///    Final value be converted into [`InputAction::Output`] using [`ActionValue::convert`].
-///
-/// [`InputCondition`]: crate::input_condition::InputCondition
-/// [`InputCondition::kind`]: crate::input_condition::InputCondition::kind
-/// [`InputModifier`]: crate::input_modifier::InputModifier
-/// [`Input`]: crate::input::Input
+/// Only when this component exists on the entity will its actions be evaluated and trigger [`events`](crate::events).
```
alice-i-cecile, line 23: This sentence will read more naturally if you invert the sentence structure.

src/lib.rs
```diff
@@ -1,38 +1,286 @@
...
-[`Release`], [`Hold`], etc. You can also add your own by implementing [`InputCondition`].
+let mut app = App::new();
+app.add_plugins((MinimalPlugins, EnhancedInputPlugin));
+```
 
-# Quick start
+## Defining actions
 
-We provide a [`prelude`] module, which exports most of the typically used traits and types.
+Actions represented by unit structs that implement the [`InputAction`] trait, which also requires
```
alice-i-cecile, line 24:
```suggestion
Actions are represented by unit structs that implement the [`InputAction`] trait, which also requires
```

src/lib.rs
```diff
@@ -1,38 +1,286 @@
...
+# use bevy_enhanced_input::prelude::*;
+# let mut app = App::new();
+# app.add_plugins(EnhancedInputPlugin);
+app.add_input_context::<OnFoot>();
+
+#[derive(InputContext)]
+struct OnFoot;
+```
+
+No required parameters needed. See the [`InputContext`] documentation for all parameters.
```
alice-i-cecile, line 61: This would be more helpfully framed as "you can provide additional parameters..."

src/lib.rs
```diff
@@ -1,38 +1,286 @@
...
+
+## Binding actions
+
+While input contexts are defined statically at compile time, bindings must be assigned at runtime. They're stored inside the
```
alice-i-cecile, line 65: You need to open with "what are input bindings".

src/lib.rs
```diff
@@ -1,38 +1,286 @@
...
+// The action will trigger when space or gamepad south button is pressed.
+actions.bind::<Jump>().to((KeyCode::Space, GamepadButton::South));
+# #[derive(InputContext)]
+# struct OnFoot;
+# #[derive(Debug, InputAction)]
+# #[input_action(output = bool)]
+# struct Jump;
+```
+
+### Input modifiers
```
alice-i-cecile, line 86: Framing and flow of this section is quite odd. I'm not immediately sure what to suggest.

alice-i-cecile (commented):
Very much an improvement, but there are a few grammatical issues to clean up.

Beyond that, I think that the framing and flow of the README is quite odd. It needs to assume less background knowledge, and clearly walk users through the data flow of doing basic things. Then, layer on complications like contexts and input modifiers on top, clearly explaining what these things are and why they're useful.

That said, I would be fine with doing minor cleanup and merging, and spinning that README refactor out to follow-up (I can do it!).

src/action_binding.rs
Shatur: Let's remove this completely, I will include this information into the evaluation flow below.

Shatur:
Thanks a lot, good points. I've pushed the changes. Let me know if anything else needs adjustment.  

> Beyond that, I think that the framing and flow of the README is quite odd. It needs to assume less background knowledge, and clearly walk users through the data flow of doing basic things.

I agree! But small suggestion: let's keep the README simple. Just outline what the crate does and present its features. I think it's better to keep API explanations in the docs with all the links and compile-checked examples.

PR #74, opened 2025-04-07
Replace UI features with `ActionSources` resource

Shatur:
Users now need to manually disable inputs via `ActionSources`.
But it’s trivial to implement, works with any UI library, and much more flexible.

I also replaced `ui_priority` with a more concise doc example.


alice-i-cecile (approved):
Lovely!

PR #78, opened 2025-04-13
Add action-level modifiers to doc examples

Shatur:


src/lib.rs
```diff
@@ -109,18 +109,25 @@ for assigning multiple modifiers at once.
...
-    KeyCode::KeyD,
-    // In Bevy sticks split by axes and captured as 1-dimensional inputs,
-    // so Y stick needs to be sweezled into Y axis.
-    GamepadAxis::RightStickX,
-    GamepadAxis::RightStickY.with_modifiers(SwizzleAxis::YXZ),
-));
+actions
+    .bind::<Move>()
+    .to((
+        // Keyboard keys captured as `bool`, but the output of `Move` is defined as `Vec2`,
```
alice-i-cecile, line 115:
```suggestion
        // Keyboard keys are captured as `bool`, but the output of `Move` is defined as `Vec2`,
```

alice-i-cecile (approved):

PR #85, opened 2025-04-15
Log read value

Shatur:
Right now we log modifier changes. But if there are no modifiers, the original value won't be logged. Let's log it.


alice-i-cecile (approved):
We might want to figure out a feature flagging system for these trace-level logs. I'm not sure if they have perf overhead when disabled.

PR #89, opened 2025-04-24
3 dimension keycode preset

Aceeri:
Adds a preset for 3 dimensional input actions.

Open to better names, SixDOF is just what came to mind for this but it sounds a bit odd.

I did a little builder pattern for wasd/arrow vs up down mappings since people might want to mix and match them. If there are any other up/down presets you think would be common lemme know :)


src/preset.rs
```diff
@@ -234,3 +234,161 @@ impl<I: IntoBindings> IntoBindings for Bidirectional<I> {
...
+///
+/// #[derive(InputContext)]
+/// struct FlyCamera;
+///
+/// #[derive(Debug, InputAction)]
+/// #[input_action(output = Vec3)]
+/// struct Move;
+/// ```
+#[derive(Debug, Clone, Copy)]
+pub struct SixDOF<I: IntoBindings> {
```
Shatur, line 289: What do you think about `Spatial` or `Tertial`?

src/preset.rs
```diff
@@ -234,3 +234,161 @@ impl<I: IntoBindings> IntoBindings for Bidirectional<I> {
         positive.chain(negative)
     }
 }
+
+/// A preset to map buttons as 3-dimensional input.
+///
+/// Uses [`SwizzleAxis`] and [`Negate`] to bind inputs to the Y and Z directions.
+///
+/// See also [`Axial`] and [`Bidirectional`].
```
Shatur, line 242: Let's point to more related presets.
```suggestion
/// See also [`Cardinal`] and [`Bidirectional`].
```

Shatur (changes requested):
Thanks you!

- Could you also add reference in `Cardinal` docs to `SixDOF` and edit `Axial` to point to only `Cardinal`?
- Ccould you also add an integration test in `presets.rs`?

src/preset.rs
```diff
@@ -234,3 +234,161 @@ impl<I: IntoBindings> IntoBindings for Bidirectional<I> {
...
+}
+
+impl SixDOF<KeyCode> {
+    /// Maps WASD keys as horizontal inputs.
+    ///
+    /// Defaults for space for up and left control for down.
+    /// See [`Self::qe`] and [`Self::space_lshift`] for other up/down presets.
+    ///
+    /// See also [`Self::arrow_keys`].
+    pub fn wasd() -> Self {
```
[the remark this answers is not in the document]
Shatur: Another option is to provide `wasd_and(up, down)` and `arrows_and(up, down)`. Still not sure, but also a thing to consider.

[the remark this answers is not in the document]
Aceeri: How about `wasd_space_lcontrol()` and `arrows_space_lcontrol()` and then have the modifying ones stay the same?

src/preset.rs
Aceeri: Spatial maybe, but I'm not sure it is better than SixDOF...

I guess it requires a bit less knowledge since you don't need to know DOF = Degrees of Freedom and what that means.

Shatur: @alice-i-cecile what do you think?

Aceeri:
>     * Ccould you also add an integration test in `presets.rs`?

Do you think I should add a new file for the 3 dimensions or modify the existing stuff to just be `Vec3`?

src/preset.rs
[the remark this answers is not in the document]
Shatur: I like the explicitness of `*_and(up, down)` more :thinking:
In real games you load these data from settings anyway.

Shatur:
I would just modify the action output and constants.

src/preset.rs
Shatur: If we decide to keep `SixDOF`, maybe worth to rename it to `SixDof` instead. The convention is to capitalize only the first letter in abbreviations - like `Aabb`.

Another option to consider: name these presets `<Something>1D`, `<Something>2D` and `<Something>3D`. Just brainstorming :slightly_smiling_face:

Shatur:
Or maybe even remove the constants in the test and use the constants from glam itself.

src/preset.rs
Aceeri: Ya maybe `Axis1D`/`Axis2D`/`Axis3D`?

src/preset.rs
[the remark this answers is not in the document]
Aceeri: Ya, *_and(up, down) sounds good to me :)

src/preset.rs
Aceeri: I guess that might make axial vs axis2d confusing... Directional*D?

Shatur (approved):
Everything looks good! But I'd like to get @alice-i-cecile's opinion on the naming. Here are the options we've considered:
- `SixDof` (renamed from `SixDOF` since in abbreviations only the first letter should be capitalized)
- `Spatial`
- `Directional*D` for all presets

If you have other suggestions, please let us know!

src/preset.rs
alice-i-cecile: I think I like "Spatial" best :) SixDoF is really confusing to new gamedevs.

alice-i-cecile (commented):
I like this addition (it was in LWIM for a reason!), but we should rename first.

Shatur:
@alice-i-cecile what do you think about [the proposed names](https://github.com/projectharmonia/bevy_enhanced_input/pull/89#pullrequestreview-2793326827)?

src/preset.rs
Shatur: @Aceeri let's go with `Spatial` then!

Shatur (approved):

alice-i-cecile (approved):

Shatur:
Never mind - resolved myself :slightly_smiling_face:

PR #90, opened 2025-04-26
Rename `Press` to `Down` and `JustPress` to `Press`

Shatur:
Reverts to Unreal Engine's original naming convention.
The names were initially changed to avoid conflicts with `bevy_picking`, but we can now reclaim them since `bevy_picking` has migrated from `Down` to `Pressed`.


alice-i-cecile (approved):
I'm down with this, but we'll need to rename upstream terms in `bevy_input` for consistency in the future.

PR #95, opened 2025-04-28
Remove uses of the word "Dummy"

Shatur:
- Use `Player` in examples. More natural.
- Use `Test` in tests. Having a context called `Test` reads great.


alice-i-cecile (approved):
I think this wording is clearer :)

janhohenheim:
Thanks, I appreciate it :D

PR #96, opened 2025-04-30
Improve actions API

Shatur:
- Follow Bevy's shift away from panicking methods.
- Rename `Actions::action` into `Actions::get`.
- Add helper methods for convenient action data access.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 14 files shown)


Shatur:
In our case it's one kind of error :thinking: 
Would you prefer to create something like `struct ActionNotFound<A, C> { action: PhantomData<A>, context: PhantomData<A> }`?

alice-i-cecile:
Yes please! Tiny and specific error types like that are surprisingly pleasant to work with: they make possible failure modes extremely clear when reading method signatures.

alice-i-cecile (approved):

PR #100, opened 2025-05-08
add time delta for Fired/Ongoing events

mockersf:
some input reaction are dependent on the duration since they were last triggered.

adds the deltas information to `Fired` and `Ongoing` events

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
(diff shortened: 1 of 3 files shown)


Shatur:
Makes sense! Could you also update the changelog?

Shatur:
Wait, can't you get this information from the `Time` resource inside your system or a trigger? :thinking:

mockersf:
yup, it's possible, but it means you now have to add also a `Local<f32>` and update it, reset it when you're changing of event sequence, ...

it's error prone and more complex to do it on consumer side

mockersf:
🤔 maybe I misunderstood and you can just use frame delta without worrying about sequence start / stop, and events are always one and only one per frame?

PR #102, opened 2025-05-18
Fix `ActionSources` example

mbillingr:
Hi, sorry to barge in with a pull request right away. This is such a small fix, I think it's hardly worth to write an issue first. If I'm wrong or you don't want my patch not much work will be wasted :)

Anyway, I was super excited to discover the `ActionSources` struct, but copying the example into my project broke mouse input. Looks like the condition is the wrong way around. Actually, we want to use the mouse for actions when there are no GUI interactions, right?

Cheers


Shatur:
Thanks!

Shatur (approved):

PR #103, opened 2025-05-21
Add new() for ConditionTimer

baozaolaoba-top:


Shatur:
But why initialize the elapsed time as non-zero? :thinking:

Shatur:
Ah, you probably want to initialize actions with a cooldown, makes sense.

Shatur (approved):
Made a small change to make the constructor a bit more explicit.

PR #110, opened 2025-06-06
Implement action mocking

Shatur:
Adds statically-typed input mocking to replace the raw API for setting action values.

`ActionMap::insert` was removed entirely (instead of made `pub(crate)`) since it was only used in tests.

This feature is useful for networking, replays, and tests.


alice-i-cecile (approved):
Also useful for bots! Really good to see this.

alice-i-cecile:
CI is red though :)

PR #112, opened 2025-06-06
Add Ordinal preset

Shatur:
I added `Ordinal` preset, which is useful for roguelikes. Each commit is a separate but related change - I recommend reviewing them one by one.


alice-i-cecile (commented):
Can we use the bevy_math cardinal directions in here somehow? It feels silly to have to reinvent this from scratch.

alice-i-cecile:
No, like [CompassOctant](https://docs.rs/bevy/latest/bevy/math/enum.CompassOctant.html) :) I don't see any way to incorporate those nicely here though, so I'll approve and we can refactor later if we see a better way.

alice-i-cecile (approved):

PR #114, opened 2025-06-07
Sort actions by the number of modifiers

Shatur:
It's common for users to create bindings in arbitrary order and expect combinations like `Ctrl + C` to take priority over just `C`. It's also common to have dynamically defined bindings where the ordering isn't known ahead of time.

Bindings are now automatically sorted before each update if needed. I had to move binding creation into a dedicated method to avoid triggering sorting during mocking.

This PR consists of two commits: a small test refactor that was needed, and the actual implementation. I recommend reviewing them separately.

### DIFF
--- a/src/action_binding.rs
+++ b/src/action_binding.rs
@@ -429,6 +429,14 @@ impl ActionBinding {
     pub(crate) fn require_reset(&self) -> bool {
         self.require_reset
     }
+
+    pub(crate) fn max_mod_keys(&self) -> usize {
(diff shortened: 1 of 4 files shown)


Tsudico:
It looks like this only applies to keyboard modifiers for now, correct? Sorting for other inputs like gamepads or joysticks that use chord combinations isn't included.

alice-i-cecile (approved):
That's a clever, fast approach to clashing input handling. I like it!

PR #117, opened 2025-06-16
Triggers rename

Shatur:
Looks like Bevy is going for short names (`Trigger` -> `On`), and I think we should do the same. I like these names more 🙂

I’m also considering splitting `Bindings` from `Actions`, so it’s better to have distinctive names.


alice-i-cecile (approved):
Agreed, I like these too!

PR #120, opened 2025-06-17
Fix time when used inside `FixedMain`

Shatur:
Previously, we used `Time<Virtual>`, but this resulted in incorrect time for contexts evaluated inside `FixedMain`.

To solve this, I introduced the `InputTime` system parameter, which holds `Time` (the default time based on the current schedule) and `Time<Real>` for non-dilated time access. This also allows us to use Bevy’s `Timer`, which is more flexible.

System parameters are a bit verbose to spawn in tests, so I had to create a small helper.


src/input_time.rs
```diff
@@ -0,0 +1,54 @@
...
+        }
+    }
+}
+
+/// Type of time resource to use.
+#[derive(Debug, Clone, Copy, Default)]
+pub enum TimeKind {
+    /// Corresponds to [`Time`].
+    ///
+    /// Virtual game time, affected by [`Time::pause`] and [`Time::relative_speed`].
```
alice-i-cecile, line 35: This should have a note about Time<Fixed>

alice-i-cecile (approved):

PR #122, opened 2025-06-18
Codebase refactor

Shatur:
Small refactoring to better organize the codebase and prepare for the rollback support for networking. I'd recommend to review commit by commit.


alice-i-cecile (approved):
These seem like reasonable refactors. I've checked the commit history and there's nothing I'd object to.

PR #123, opened 2025-06-19
Make pull-based API more strongly typed

Shatur:
We already return the exact action's output type from `Actions::value`, but `Actions::get` still returns an `Action` that holds an `ActionValue`.

Renamed `Action` to `UntypedAction` and introduced a new `Action` type that is generic and holds the action's output type. It can be constructed from an `UntypedAction`. This allows returning a strongly typed action from `Actions::get`.


alice-i-cecile (approved):
Give me types! More types!

PR #125, opened 2025-06-20
Minor changes for networking

cBournhonesque:


src/input_context/actions.rs
```diff
@@ -141,10 +141,15 @@ impl<C: InputContext> Actions<C> {
...
-    /// Returns an mutable iterator over type-erased actions data and their IDs.
+    /// Returns a mutable iterator over type-erased actions data and their IDs.
     pub fn iter_mut(&mut self) -> impl Iterator<Item = (TypeId, &mut UntypedAction)> {
         self.action_map.iter_mut().map(|(&id, action)| (id, action))
     }
+    
+    /// Returns the UntypedAction for the given action type if it exists.
+    pub fn get_mut(&mut self, action_type: TypeId) -> Option<&mut UntypedAction> {
+        self.action_map.get_mut(&action_type)
+    }
```
Shatur, line 152: Let's rename the function a little since we have a regular typed `get`:
```suggestion 
    /// Returns a type-erased action for the given type ID if it exists.
    pub fn get_mut_by_id(&mut self, type_id: TypeId) -> Option<&mut UntypedAction> {
        self.action_map.get_mut(&type_id)
    }
```

Shatur (approved):

src/input_context/actions.rs
```diff
@@ -141,11 +141,16 @@ impl<C: InputContext> Actions<C> {
...
+    /// Returns a mutable iterator over type-erased actions data and their IDs.
     pub fn iter_mut(&mut self) -> impl Iterator<Item = (TypeId, &mut UntypedAction)> {
         self.action_map.iter_mut().map(|(&id, action)| (id, action))
     }
 
+    /// Returns the UntypedAction for the given action type if it exists.
+    pub fn get_mut(&mut self, action_type: TypeId) -> Option<&mut UntypedAction> {
+        self.action_map.get_mut(&action_type)
+    }
+
```
Shatur, line 153: Looks like the proposed change got reverted. It's probably by accident since you applied the change from the browser and then force-pushed the formatting.
```suggestion
    /// Returns a type-erased action for the given type ID if it exists.
    pub fn get_mut_by_id(&mut self, type_id: TypeId) -> Option<&mut UntypedAction> {
        self.action_map.get_mut(&type_id)
    }
```

Shatur (changes requested):

Shatur (approved):

PR #128, opened 2025-06-22
Updates for networking

cBournhonesque:
- On the receiver (server), i need to be able to bind an action to a context in an untyped manner. This change allows me to register a function pointer for each `InputAction`, instead of for each `InputAction X InputContext`
- I need to have a bit more flexibility while ordering systems, I would like to be able to schedule systems between `BEI::Update` and `BEI::Trigger`, so I added a new system set for Trigger
- I decided to network the ActionEvents as well, so I need a Serialize/Desreialize implementation on them

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - `Action<A>` now implements `Clone` and `Copy` for any `A`.
+- `ActionEvents` now implements `Serialize` and `Deserialize`.
+- `BEIEnhancedSystem` has been split between `BEIEnhancedSet::Update` (read new inputs from the `InputReader` and update the `Actions` components)
+ and `BEIEnhancedSet::Trigger` (trigger the events corresponding to how the `Actions` components changed)
+- Most of the `Actions<C>` functionalities have been moved to an untyped struct `UntypedActions`. `Actions<C>` derefs to `UntypedActions` so you
+don't have to change any call site.
 
 ## [0.13.0] - 2025-06-19
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `Action<A>` now implements `Clone` and `Copy` for any `A`.
 - `ActionEvents` now implements `Serialize` and `Deserialize`.
-- `BEIEnhancedSystem` has been split between `BEIEnhancedSet::Update` (read new inputs from the `InputReader` and update the `Actions` components)
- and `BEIEnhancedSet::Trigger` (trigger the events corresponding to how the `Actions` components changed)
+- `EnhancedInputSystem` has been split between `EnhancedInputSet::Update` (read new inputs from the `InputReader` and update the `Actions` components)
+ and `EnhancedInputSet::Trigger` (trigger the events corresponding to how the `Actions` components changed)
 - Most of the `Actions<C>` functionalities have been moved to an untyped struct `UntypedActions`. `Actions<C>` derefs to `UntypedActions` so you
 don't have to change any call site.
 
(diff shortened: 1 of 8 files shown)


src/input_context/actions.rs
```diff
@@ -24,8 +25,8 @@ use crate::{
 #[derive(Component)]
 pub struct Actions<C: InputContext> {
     gamepad: GamepadDevice,
-    bindings: Vec<ActionBinding>,
-    action_map: TypeIdMap<UntypedAction>,
+    pub bindings: Vec<ActionBinding>,
+    pub action_map: TypeIdMap<UntypedAction>,
```
Shatur, line 29: These two fields needs to be kept in sync, so we need to encapsulate them. I played with the API for it locally a little and I'd suggest the following:
- Remove `C` from `NoActionError`. I think it will be obvious for the user from which context the error is from.
- Then move all fields except `marker` from `Actions<C>` to `UntypedActions`.
- Store `UntypedActions` inside `Actions<C>` and deref to `UntypedActions`.

Maybe as a separate PR to simplify this one? :thinking:

Shatur (changes requested):

Shatur (changes requested):
Left a few suggestions.

Splitting updating and triggering is a good change.
Not quite satisfied with type-erasure of `Actions<C>` since we now need to point to `UntypedActions` everywhere in our docs, while users utilizes `Actions<C>`. But I can't suggest anything better for now, so let's unblock @cBournhonesque and draft a new release after it.

I currently experimenting with component-based design for actions. If I succeed, the type erasure won't be needed in the future.

src/input_context/actions.rs
```diff
@@ -17,20 +17,42 @@ use crate::{
...
+    fn deref(&self) -> &Self::Target {
+        &self.actions
+    }
+}
+
+impl<C: InputContext> DerefMut for Actions<C> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.actions
+    }
+}
```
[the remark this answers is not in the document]
cBournhonesque: This doesn't work because of the `C: InputContext` bound; which would mean that the Deref/DerefMut are derived only if C implements Deref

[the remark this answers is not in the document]
cBournhonesque: oh you're right, when i tried it i must still have been using the core Deref

src/input_context/actions.rs
```diff
@@ -6,9 +6,9 @@ use core::{
     fmt::{self, Debug, Display, Formatter},
     marker::PhantomData,
 };
+use log::debug;
 
 use bevy::{platform::collections::hash_map::Entry, prelude::*, utils::TypeIdMap};
-use log::debug;
```
Shatur, line 11: Please, align the log as before. Can't make suggestions since git thinks it's a deleted line.

Shatur (changes requested):
One tiny thing left :slightly_smiling_face:

src/input_context/actions.rs
cBournhonesque: oops!

CHANGELOG.md
```diff
@@ -10,6 +10,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - `Action<A>` now implements `Clone` and `Copy` for any `A`.
+- `ActionEvents` now implements `Serialize` and `Deserialize`.
+- `BEIEnhancedSystem` has been split between `BEIEnhancedSet::Update` (read new inputs from the `InputReader` and update the `Actions` components)
+ and `BEIEnhancedSet::Trigger` (trigger the events corresponding to how the `Actions` components changed)
+- Most of the `Actions<C>` functionalities have been moved to an untyped struct `UntypedActions`. `Actions<C>` derefs to `UntypedActions` so you
+don't have to change any call site.
```
Shatur, line 17:
```suggestion
- `ActionEvents` now implements `Serialize` and `Deserialize`.
- `EnhancedInputSystem` has been split between `EnhancedInputSet::Update` (read new inputs from the `InputReader` and update the `Actions` components)
 and `EnhancedInputSet::Trigger` (trigger the events corresponding to how the `Actions` components changed)
- Most of the `Actions<C>` functionalities have been moved to an untyped struct `UntypedActions`. `Actions<C>` derefs to `UntypedActions` so you
don't have to change any call site.
```

alice-i-cecile (approved):

PR #132, opened 2025-06-24
Consume inputs between contexts

Shatur:
I noticed that input consumption was local to the context's schedule, while reset inputs were global. Now both are global, as this is more expected and convenient.


alice-i-cecile (approved):

Shatur:
I should have run all tests. They fail because this approach won't work properly with a fixed schedule. Turning into a draft for now.

Issue #133, opened 2025-06-25
`Pulse` condition triggers `Fired` only on the first actuation


alecrabbit:
**Expected**
Using the `Pulse` condition on an input should trigger `Fired` on every new actuation (i.e., whenever the input is pressed after being released).

**Actual**
Gives `Fired` only on **first** actuation, and not on subsequent presses.

**How to reproduce**
1. Run the examples/all_conditions.rs example.
2. Interact with the input that uses the Pulse condition.
3. Observe that only the first press triggers the Fired event.

PR #134, opened 2025-06-25
Fix for #133 (Pulse condition triggers Fired only on the first actuation)

alecrabbit:
Pulse condition triggers Fired only on the first actuation #133

### DIFF
--- a/src/input_context/input_condition/pulse.rs
+++ b/src/input_context/input_condition/pulse.rs
@@ -154,7 +154,7 @@ mod tests {
     }
 
     #[test]
-    fn pulse_fires_again_after_release() {
+    fn fires_again_after_release() {
         let mut condition = Pulse::new(1.0);
         let action_map = TypeIdMap::<UntypedAction>::default();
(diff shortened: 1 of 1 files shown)


src/input_context/input_condition/pulse.rs
```diff
@@ -152,6 +153,45 @@ mod tests {
         );
     }
 
+    #[test]
+    fn pulse_fires_again_after_release() {
```
Shatur, line 157: Since it's inside `pulse` module, I prefer to shorten test names. Yes, we have a `pulse` test above, but I couldn't come up with a nicer name for it :sweat_smile: 
```suggestion
    fn fires_again_after_release() {
```

Shatur (approved):

src/input_context/input_condition/pulse.rs
alecrabbit: ah, naming... 🫠
you are the boss 😂

alice-i-cecile (approved):

PR #137, opened 2025-06-25
Fix input consumption between different schedules

Shatur:
Previously, input consumption was local to a context’s schedule, while reset inputs were global. Now both are global, and inputs are consumed properly across different schedules.

Supersedes #132. This implementation tracks which schedule each input was consumed in and resets them only when that schedule runs. Supports both fixed and non-fixed schedules.

### DIFF
--- a/src/input_reader.rs
+++ b/src/input_reader.rs
@@ -1,14 +1,20 @@
 use alloc::vec::Vec;
-use core::hash::Hash;
+use core::{any::TypeId, hash::Hash, iter, mem};
 
 use bevy::{
-    ecs::system::SystemParam,
+    ecs::{schedule::ScheduleLabel, system::SystemParam},
     input::mouse::{AccumulatedMouseMotion, AccumulatedMouseScroll},
     platform::collections::HashSet,
     prelude::*,
+    utils::TypeIdMap,
 };
+use log::trace;
 
 use crate::prelude::*;
+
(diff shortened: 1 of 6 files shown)


src/input_reader.rs
```diff
@@ -258,9 +265,30 @@ impl Default for ActionSources {
     }
 }
 
-/// Tracks all consumed input from Bevy resources.
+/// All consumed input by actions in each schedule.
```
alice-i-cecile, line 268: Brief context of why this exists would be really helpful to record here.

src/input_reader.rs
```diff
@@ -258,9 +265,30 @@ impl Default for ActionSources {
...
+#[derive(Resource, Default, Deref, DerefMut)]
+pub(crate) struct ConsumedInputs(TypeIdMap<IgnoredInputs>);
+
+/// Inputs that will be ignored until they return zero.
+///
+/// Once the input becomes zero, it will be automatically removed and no longer ignored.
```
alice-i-cecile, line 274: It would be nice to document (or link to documentation) about what "zero" means for the various kinds of input. Or choose a different term?

src/input_reader.rs
Shatur: This and the ones below are private structs, but I agree we should document them since the logic is quite tricky.

Issue #129, opened 2025-06-23
Component-based API


Shatur:
I have been thinking for component-based API and I think with the current Bevy state we can actually try it.

Here is my initial vision for the component-based API:
- `InputContext` now additionally requires `Component`.
  - It’s a trait instead of a struct because we want to support multiple contexts on the same entity. This is also why we need to register contexts - to know which types to evaluate.
  - To instantiate a context, the user simply inserts it as a component.
- Actions for contexts are represented by entities with an `ActionOf<C>` relation, with `Actions<C>` as the target located on the context entity (which now only stores entities).
  - Each action must have an `Action` component.
    - It’s not generic, but its construction method is: `Action::new::<C: InputAction>()`. This allows us to store a function pointer to trigger events for this action.
    - Requires: `Name`, `ActionState`, `ActionValue`, `ActionEvents`, and `ActionTime` (timing information) components. Making them granular is more convenient for networking and plays better with change detection. We previously stored this data inside `UntypedAction`.
  - Can optionally contain a `Modifiers` component, which stores `Vec<Box<dyn InputModifier>>`.
  - Can optionally contain a `Conditions` component, which stores `Vec<Box<dyn InputCondition>>`.
  - Can optionally contain a `Mock` component that overrides bindings.
- Bindings are represented by entities with `BindingOf<A>` relations, with `Bindings<C>` as the target located on an action entity.
  - Each binding contains an `Input` component with the associated input.
  - Can optionally contain a `Modifiers` component.
  - Can optionally contain a `Conditions` component.
- Presets are now bundles.

**Pros:**
- Easier to network.
- Can be represented as a scene, which is a big plus for the future editor.
- Reuses existing Bevy API instead of a custom-implemented tuple-like solution.
- Users can query actions for a pull-based approach.

**Cons:**
- The pull-based approach is no longer fully typed and uses the `ActionValue` enum instead. But we still have a fully typed push-based API based on triggers.
- Context evaluations will be slower since we now perform lookups. But users usually don’t have a ton of actions.

**Open questions:**
- Maybe we could think of a better API for modifiers and conditions as well? Currently, it’s unchanged.

alice-i-cecile:
I'm quite happy with this direction: I think it'll be more flexible and much easier to debug.

Shatur:
Duplicating my thoughts from Discord.
Modifiers and conditions in their current form don’t work nicely with this approach. Two problems:
[...]

PR #140, opened 2025-07-17
Rewrite into component-based API

### LINKED ISSUE #101: Load/Save Bindings Support using Bevy AssetLoader/Transformer/Saver
Incorporating Bevy's asset management into your plugin would allow input bindings to be loaded/saved using the existing infrastructure provided by Bevy's AssetServer.

https://docs.rs/bevy/latest/bevy/asset/index.html
https://docs.rs/bevy/latest/bevy/asset/trait.AssetLoader.html
https://docs.rs/bevy/latest/bevy/asset/saver/index.html

Shatur:
This PR implements #129, but some aspects are a bit different from what I initially described in the issue (in a better way!):

- Input contexts now components.
- Actions for contexts now represented by entities with an `ActionOf<C>` relation, where `Actions<C>` is the target located on the context entity (which now only stores entities).
- `Action` now holds only the action value and its fields, with `ActionState`, `ActionValue`, and `ActionEvents` now being components. Timing is now stored in a new `ActionTime` component.
- `InputAction` now only contains the associated `Output` type. All action settings are now expressed by the `ActionSettings` component, which can be modified at runtime.
- Bindings for actions now represented by entities with `BindingOf<A>` relations, where `Bindings<C>` is the target located on an action entity.
- Rename `Input` into `Binding` which now a component that represents the assigned binding.
- Modifiers and conditions now regular components on action and binding entities. Custom modifiers and conditions now needs to be registered using `InputModifierAppExt::add_input_modifier` and `InputConditionAppExt::add_input_condition` respectively.
- Mocking now represented by `ActionMock` component that can be added to action entities.
- Presets now represented by `SpawnableList`s and store bundles. To assign multiple items as before, just spawn multiple presets. For empty bindings inside presets we now provide convenient `Binding::None`.

I didn't expect to make that many changes, especially since the concepts are the same - they're just represented differently. But I had to rewrite a lot of docs, adjust all API usages in tests and examples, and reorganize everything. However, I have high confidence in the changes, as all tests pass and examples work as expected.

Pros of the new API:
- You can use Bevy scenes to describe bindings. When BSN lands, we'll get hot-reloadable actions for free. In UE, this is also expressed via Blueprints. This basically closes #101.
- Everything is inspectable.
- Modifiers and conditions are now accessible in systems, allowing you to adjust or read their data if needed.
- Actions can be queried - especially convenient when using `Single`.
- Things that didn't benefit from being static, such as prioritization and action settings, are now dynamic.
- Things that were "inspired by Bevy," such as how bindings were defined, modifiers and presets, now actually utilize the Bevy API - meaning new users don't have to learn additional concepts.
- The crate logic is much simpler.

Cons:
- Some things are a bit more verbose due to how hierarchy spawning works in Bevy. But this could be improved by Bevy itself. I have high hopes for BSN.
- The `bindings!` macro doesn’t allow mixing single elements and tuples (see its documentation). We could solve this by switching to a proc macro, but I don't think it's worth it. Or if you have an idea how we can adjust macro rules - let me know :slightly_smiling_face: 

API showcase:

```rust
// Spawn a camera with an input context.
commands.spawn((
    Camera3d::default(),
    Transform::from_xyz(-2.5, 4.5, 9.0).looking_at(Vec3::ZERO, Vec3::Y),
    FlyCam,
    // Similar to `related!`, but you only specify the context type.
    // Actions are related to specific context since a single entity can have multiple contexts.
    actions!(FlyCam[
        (
            Action::<Move>::new(),
            // Conditions are components.
            DeadZone::default(),
            SmoothNudge::default(),
            // Modifiers also components.
            Scale::splat(0.3),
            // Bindings are entities related to actions.
            Bindings::spawn((Cardinal::wasd_keys(), Axial::left_stick())),
        ),
        (
            Action::<Rotate>::new(),
            Bindings::spawn((
                // You can attach modifiers to individual bindings as well.
                Spawn((Binding::mouse_motion(), Scale::splat(0.1), Negate::all())),
                Axial::right_stick().with((Scale::splat(2.0), Negate::x())),
            )),
        ),
        // For bindings we also have a macro similar to `children!`.
        (Action::<CaptureCursor>::new(), bindings![MouseButton::Left]),
        (Action::<ReleaseCursor>::new(), bindings![KeyCode::Escape]),
    ]),
));
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -313,6 +313,41 @@ fn apply_input(
 # struct Move;
 ```
 
+## Removing contexts
+
+To remove a context from an entity, you need to remove it with required components **and** despawn its actions.
+
+```
+# use bevy::prelude::*;
+# use bevy_enhanced_input::prelude::*;
+# let mut world = World::new();
+let mut player = world.spawn((
+    OnFoot,
+    actions!(OnFoot[
+        (Action::<Jump>::new(), bindings![KeyCode::Space, GamepadButton::South]),
+        (Action::<Fire>::new(), bindings![MouseButton::Left, GamepadButton::RightTrigger2]),
+    ])
+));
+
+player
+    .remove_with_requires::<OnFoot>()
+    .despawn_related::<Actions<OnFoot>>();
+
+assert_eq!(world.entities().len(), 1, "only the player entity should be left");
+# #[derive(Component)]
+# struct OnFoot;
+# #[derive(InputAction)]
+# #[action_output(bool)]
+# struct Jump;
+# #[derive(InputAction)]
+# #[action_output(bool)]
+# struct Fire;
+```
+
+Actions aren't despawned automatically via [`EntityWorldMut::remove_with_requires`], since Bevy doesn't automatically
+despawn related entities when their relationship targets (like [`Actions<C>`]) are removed. For this reason, [`Actions<C>`]
+is not a required component for `C`. See [this issue](https://github.com/bevyengine/bevy/issues/20252) for more details.
+
 ## Input and UI
 
 Currently, we don't integrate `bevy_input_focus` directly. But we provide [`ActionSources`] resource


janhohenheim (approved):
No comment on the implementation, but looks solid from a user perspective :)

Shatur:
Re-read all the code and made a few final touches, mostly to the docs and examples 🙂  
I also reviewed the coverage report and added the necessary tests. The remaining uncovered lines are false positives. Rust's coverage tooling definitely needs some love.

CHANGELOG.md
```diff
@@ -7,6 +7,31 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+This update features a big rewrite into a component-based API. The core concepts remain the same, but are now expressed through ECS. It's recommended to revisit the quick start guide.
+
+### Changed
+
+- Input contexts are now components.
+- Actions for contexts are now represented by entities with an `ActionOf<C>` relation, where `Actions<C>` is the target located on the context entity (which now only stores entities).
```
alice-i-cecile, line 15: This sentence isn't clear to me as a reader. "is the target" is particularly unclear to me, and it would be nice to introduce "context entities" in advance.

src/binding.rs
```diff
@@ -0,0 +1,238 @@
...
+        }
+    }
+}
+
+fn reset_first_activation(mut world: DeferredWorld, ctx: HookContext) {
+    let mut first_activation = world.get_mut::<FirstActivation>(ctx.entity).unwrap();
+    **first_activation = true;
+}
+
+/// Whether the input output a non-zero value.
```
alice-i-cecile, line 174: I don't understand what this sentence means at all.

src/binding.rs
```diff
@@ -0,0 +1,238 @@
...
+    }
+}
+
+impl From<GamepadAxis> for Binding {
+    fn from(value: GamepadAxis) -> Self {
+        Self::GamepadAxis(value)
+    }
+}
+
+/// A trait to ergonomically assign keyboard modifiers to any type that can be converted into a binding.
```
alice-i-cecile, line 143:
```suggestion
/// A trait to ergonomically assign keyboard modifiers to any type that can be converted into a [`Binding`].
```

src/binding.rs
```diff
@@ -0,0 +1,238 @@
...
+
+use bevy::{
+    ecs::{component::HookContext, world::DeferredWorld},
+    prelude::*,
+};
+use serde::{Deserialize, Serialize};
+
+use crate::prelude::*;
+
+/// Input from any source associated with [`Action<C>`].
```
alice-i-cecile, line 14: This is a weird definition. A binding to me is "the mapping between actions and inputs". This definition suggests it... statefulyl tracks the current input?

src/action.rs
```diff
@@ -0,0 +1,405 @@
...
+    ///
+    /// - Use [`bool`] for button-like actions (e.g., `Jump`).
+    /// - Use [`f32`] for single-axis actions (e.g., `Zoom`).
+    /// - For multi-axis actions, like `Move`, use [`Vec2`] or [`Vec3`].
+    type Output: ActionOutput;
+}
+
+/// Marks a type which can be used as [`InputAction::Output`].
+pub trait ActionOutput: Into<ActionValue> + Default + Send + Sync + Debug + Clone + Copy {
+    /// Dimension of this output.
```
alice-i-cecile, line 120: Some examples would be helpful here.

src/action.rs
```diff
@@ -0,0 +1,405 @@
...
+    const DIM: ActionValueDim;
+
+    /// Converts the value into the action output type.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the value represents a different type.
+    fn unwrap_value(value: ActionValue) -> Self;
```
alice-i-cecile, line 128: Non-blocking, but I don't think this should panic. Instead, return a result and let the caller unwrapper or recover.

src/context.rs
```diff
@@ -0,0 +1,623 @@
...
+            .add_observer(register_instance::<C, S>)
+            .add_observer(remove_context::<C, S>);
+
+        self
+    }
+}
+
+/// Tracks registered input contexts for each schedule.
+///
+/// In Bevy, it’s impossible to know which schedule is used inside a system,
```
alice-i-cecile, line 109: I bet we could expose this in the future.

Shatur: Ah, it would be great! :slightly_smiling_face:

src/action.rs
Shatur: This trait is for `ActionValue` creation and its conversion to `Action<C>`. Users don't need to directly use it. But I updated the documentation for it. Let me know if it's clearer now!

src/action.rs
Shatur: Same as above - users won't use it and I used panic only because this should never happen: we create right dynamic type for an action and just unwrap it into the right static type. If the panic triggers - it's a but in my implementation :slightly_smiling_face:

Shatur (commented):
Thank you! Just pushed the changes to address the suggestions. Feel free to comment on them if something is still looks unclear!

PR #149, opened 2025-07-28
Chord Firing on any sub action activation

slyedoc:
Rework a few with new Component changes, nice work by the way.   Noticed Chord didn't work as expected, tested with all_conditions example and ChordAction fires when any of the sub actions fire.  Not sure this fix is ideal but seems fix my issue.


Shatur:
Thank you! I decided to fix it a bit differently - it’s now capped at `Ongoing` if any of the actions are not `Fired`. This should make it more intuitive and emphasize that the conditions are only partially met.

alice-i-cecile (approved):

PR #152, opened 2025-08-09
Rework `bindings!` and `actions!` macro

Shatur:
It's impossible to properly express "convert first element into Binding" with a `macro_rules`. But I realized that we can just move some of the things into a type system. This way it properly works with trailing commas and no longer requires wrapping single elements in braces when mixed with tuples.


alice-i-cecile (changes requested):
Needs regression tests :)

Shatur:
They implemented as doc tests :slightly_smiling_face:

Shatur:
Included one tiny fix with `related!`. See the commit message for details.

alice-i-cecile (approved):

PR #153, opened 2025-08-09
Remove `DeadZone::with_lower_threshold` and

Shatur:
`DeadZone::with_upper_threshold`

Cart generally prefers not to have these because they're annoying to maintain, bloat compile times, and offer a second way to do things that isn't much better. They're also not BSN compatible.

This makes sense when all fields are public. In our case, it's only `DeadZone`.

There are a few other places, but they have private fields. For those, I think convenience methods are justified, since we can't initialize them using struct initialization syntax, so there is no API duplication.

@alice-i-cecile feel free to suggest more things like this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Removed
+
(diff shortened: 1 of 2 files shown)


alice-i-cecile:
> There are a few other places, but they have private fields. For those, I think convenience methods are justified, since we can't initialize them using struct initialization syntax, so there is no API duplication.

Yep, this makes sense to me.

alice-i-cecile (approved):

PR #158, opened 2025-08-09
Add context activation/deactivation

Shatur:
Sometimes it's useful to temporarily deactivate a context without removing it. This PR introduces `ContextActivity` component, which is similar to Bevy's `Visibility`.

I split changes into 2 logical commits: first the actual implementation and the second one updates `context_swtich` example to use it.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -28,7 +28,7 @@ use crate::{
     action::fns::ActionFns,
     binding::FirstActivation,
     condition::fns::{ConditionFns, ConditionRegistry},
-    context::trigger_tracker::TriggerTracker,
+    context::{input_reader::PendingBindings, trigger_tracker::TriggerTracker},
     modifier::fns::{ModifierFns, ModifierRegistry},
     prelude::*,
 };
@@ -80,6 +80,7 @@ impl InputContextAppExt for App {
         );
 
         let actions_id = self.world_mut().register_component::<Actions<C>>();
+        let activity_id = self.world_mut().register_component::<ContextActivity<C>>();
         let mut registry = self.world_mut().resource_mut::<ContextRegistry>();
         if let Some(contexts) = registry
             .iter_mut()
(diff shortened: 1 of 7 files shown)


Shatur:
Rebased to the latest master.

Shatur:
Looks like [in Discord](https://discord.com/channels/691052431525675048/1297361733886677036/1404117459929268254) users think positively about this feature.

src/context.rs
```diff
@@ -518,6 +564,70 @@ fn apply<S: ScheduleLabel>(
...
+/// This is analogous to hiding an entity instead of despawning.
+/// Use this component when you want to toggle quickly, preserve bindings, or keep entity IDs.
+/// Use removal when the context is truly going away and you don't need it back soon.
+///
+/// Marked as required for `C` on context registration.
+#[derive(Component, Reflect, Deref)]
+#[component(immutable)]
+pub struct ContextActivity<C> {
+    #[deref]
+    value: bool,
```
alice-i-cecile, line 585: This should be called `enabled`, or something similarly semantic.

alice-i-cecile:
> And we don't just stop evaluating it, we also trigger the transition to None, so all observers also behave as expected (just like with removals).

We should be changing entity disabling to work like this upstream BTW. If that was resolved, could we use a hypothetical component disabling instead?

I still don't like adding an ad hoc mechanism for this, because I've seen this result in a ton of complexity for both users and maintainers.

Shatur:
> If that was resolved, could we use a hypothetical component disabling instead?

Yes, this would work, but I'm not sure if it would be the right tool. The idea of entity or component disabling is to hide them from queries, but we don't need to hide them - we just need to disable the evaluation logic.  
I'd look at this like visibility: we don't remove hidden entities from queries.

alice-i-cecile (approved):
Alright, I won't block this further. I may come back to refactor this later.

PR #159, opened 2025-08-09
Split `Binding::MouseWheel` into 2 variants

Shatur:
`Binding::VertWheel` and `Binding::SideWheel`. Each captures input for a different axis as `Axis1D`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - `bindings!` now properly works with trailing commas and no longer requires wrapping single elements in braces when mixed with tuples.
+- Split `Binding::MouseWheel` into `Binding::VertWheel` and `Binding::SideWheel`. Each captures input for a different axis as `Axis1D`.
 
 ### Removed
 
--- a/src/binding.rs
+++ b/src/binding.rs
@@ -33,11 +33,10 @@ pub enum Binding {
     },
     /// Mouse movement, captured as [`ActionValue::Axis2D`].
     MouseMotion { mod_keys: ModKeys },
-    /// Mouse wheel, captured as [`ActionValue::Axis2D`].
-    ///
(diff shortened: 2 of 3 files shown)


Tsudico:
I am a little concerned that `VertWheel` might be harder to find for users who are looking for the default mouse wheel behavior. Would it be better to have the same name as what users expect, such as `MouseWheel`, or if you want to have the specific scroll behavior included then `MouseVertWheel` might be a good choice so that it is easier for users to find in documentation and understand in code?

Shatur:
Good point! How about `MouseWheel` and `MouseSideWheel`?
I think it's better to use short names for "high-traffic" items, since typing `MouseVertWheel` or `mouse_vert_wheel` for fields or helpers could get annoying.

alecrabbit:
imo, it would be better for `MouseWheel` to be left as is as it's data sources: bevy's [MouseWheel](https://docs.rs/bevy/latest/bevy/input/mouse/struct.MouseWheel.html) and winit's `MouseWheel`
these have x and y values
so, I suggest for the enhacement to **add** two additional bindings like `MouseWheelVertcalOnly` and `MouseWheelHorizontalOnly` (exact naming does not matter much for me)

Shatur:
I don't like the idea of having 3 variants for a wheel. But I agree on `MouseWheel` naming. We can create a `MouseWheel` *preset* to output a `Vec2`, this would be cleaner. But then how to name the variants?

Another option is to keep `MouseWheel` variant and create a presets for only horizontal and vertical scrolling.

Shatur:
Thinking about it more, all presets are `SpawnableList`s, but for the wheel we only need to apply a Swizzle, so it's just a `Bundle`.
And nor `Bundle` or `SpawnableList` won't be directly usable for `bindings!` macro...

Maybe just better document it? We already mention that `MouseWheel` outputs `Vec2`, but maybe add an example? :thinking:

Tsudico:
A code example added to the MouseWheel documentation right under where it mentions the Swizzle might better call attention to it and show exactly what needs to be done. In addition, the simple fly cam could be enhanced to possibly use the binding to alter the movement speed of the fly cam. That way there are multiple locations for the information.

PR #160, opened 2025-08-09
Manually impl derives for `ActionOf<C>`

cBournhonesque:


alice-i-cecile (approved):

Shatur:
Thanks! Added other impls to avoid requiring bounds on `C`.

PR #163, opened 2025-08-10
Rename preset helper methods

Shatur:
`horizontal_arrow_keys` and `vertical_arrow_keys` are quite long.


alice-i-cecile (changes requested):
I think the current rename is a bit of a lateral move. The x/y terminology is confusing here. How about:

- `Bidirectional::arrow_left_right` /  `Bidirectional::arrow_up_down`
- `Bidirectional::left_right` /  `Bidirectional::up_down`

Shatur:
What do you think about these names?

The strategy:
- For keyboard keys, add the `_keys` prefix because while `wasd` is common, `hjkl` or just `ws`/`ad` look weird.
- For arrows, numpad, and dpad, it's obvious what they are, so drop `_keys`/`_buttons`.
- For bidirectional, use `up_down` and `left_right`.

alice-i-cecile (approved):

alice-i-cecile:
I like it!

PR #166, opened 2025-08-10
Don't trigger change detection on `Actions<C>` due to sorting

Shatur:
We sort actions based on their number of modifiers. There is no reliable way to detect whether sorting is needed, because bindings on actions can technically not only be inserted but also reparented.

So we sort them every time. Sorting in Rust is already optimized for unchanged collections. However, this triggers change detection every frame. Since users usually expect a relationship target to change only when its elements actually change, I bypassed the change detection as a workaround.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -302,10 +302,8 @@ fn update<S: ScheduleLabel>(
             continue;
         };
 
-        // We only sort actions, without actually modifying elements the collection.
-        let context_actions = context_actions.bypass_change_detection();
-        context_actions.sort_by_cached_key(|&action| {
-            let Ok((.., action_bindings, _, _, _)) = actions.get(action) else {
+        let mods_count = |action: &Entity| {
+            let Ok((.., action_bindings, _, _, _)) = actions.get(*action) else {
                 // TODO: use `warn_once` when `bevy_log` becomes `no_std` compatible.
                 warn!(
                     "`{action}` from `{}` missing action components",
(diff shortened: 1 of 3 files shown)


src/context.rs
```diff
@@ -298,10 +298,12 @@ fn update<S: ScheduleLabel>(
         };
 
         let gamepad = context.get::<GamepadDevice>().copied().unwrap_or_default();
-        let Some(context_actions) = instance.actions_mut(&mut context) else {
+        let Some(mut context_actions) = instance.actions_mut(&mut context) else {
             continue;
         };
 
+        // We only sort actions, without actually modifying elements the collection.
+        let context_actions = context_actions.bypass_change_detection();
```
alice-i-cecile, line 306: I would instead guard this by a `.is_sorted` call or the like.

alice-i-cecile (commented):
Can you give me a bit more context about why we're sorting actions? The existing solution is fine, but if that sorting is ever semantically meaningful to users, we should be exposing the true positives correctly instead. Which seems like it should be viable to do by checking if the list is sorted before sorting.

src/context.rs
Shatur: This might be a good idea :thinking: Let me try.

Shatur:
We sort actions by their number of modifiers. So if you have `Ctrl + A` and just `A`, we evaluate `Ctrl + A` first. This relevant for input consumption.

But your suggestion is better :slightly_smiling_face:  It's also might be even faster since sorting is rarely needed.

PR #175, opened 2025-08-16
Add `action` to events

Shatur:
All actions target an entity with a context. This makes sense, since users can attach an input context to their player and react to actions in a convenient way.
However, sometimes it is useful to know which action entity triggered the event - especially if you have multiple identical actions with different bindings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Target context entity on events triggered by action removals.
+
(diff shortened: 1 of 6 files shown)


alice-i-cecile (approved):
Yes please! Really good change.

Shatur:
@alice-i-cecile fixed the bug. It was necessary to resolve this before implementing the feature, so I included it here. I split the work into independent commits and added explanations there. Let's just merge this PR with rebase :slightly_smiling_face:

Sirmadeira (approved):

Shatur:
@alice-i-cecile Ah, it was merged because you approved it previously. Feel free to leave suggestions here, I'll open a follow-up if necessary.

PR #179, opened 2025-08-19
Add hotbar example

van-nessing:
Add an example that shows how to use the action field on events.


alice-i-cecile (approved):
Nice little usage example! It will be nice to beef this up with UI code after upstreaming.

Shatur (approved):

PR #180, opened 2025-08-19
Cooldown input condition

idangur:
Created this for my game and thought this might be useful to have in the library itself, 
Open for any changes and can close this PR if it's not needed

### DIFF
--- a/examples/all_conditions.rs
+++ b/examples/all_conditions.rs
@@ -2,7 +2,6 @@
 //! Press keys from the number row on the keyboard to trigger actions and observe the output in console.
 
 use bevy::{ecs::spawn::SpawnWith, log::LogPlugin, prelude::*};
-use bevy_enhanced_input::condition::cooldown::Cooldown;
 use bevy_enhanced_input::prelude::*;
 
 fn main() {
@@ -67,11 +66,6 @@ fn spawn(mut commands: Commands) {
                 Tap::new(0.5),
                 bindings![TestTap::KEY],
             ));
-            context.spawn((
-                Action::<TestCooldown>::new(),
-                Cooldown::new(1.0),
-                bindings![TestCooldown::KEY],
-            ));
 
 
             let member1 = context
@@ -90,6 +84,12 @@ fn spawn(mut commands: Commands) {
                 Action::<TestBlockBy>::new(),
                 BlockBy::single(blocker),
                 bindings![TestBlockBy::KEY],
+            ));
+            
+            context.spawn((
+                Action::<TestCooldown>::new(),
+                Cooldown::new(1.0),
+                bindings![TestCooldown::KEY],
             ));
         })),
     ));
--- a/examples/all_conditions.rs
+++ b/examples/all_conditions.rs
@@ -2,6 +2,7 @@
 //! Press keys from the number row on the keyboard to trigger actions and observe the output in console.
 
 use bevy::{ecs::spawn::SpawnWith, log::LogPlugin, prelude::*};
+use bevy_enhanced_input::condition::cooldown::Cooldown;
 use bevy_enhanced_input::prelude::*;
 
 fn main() {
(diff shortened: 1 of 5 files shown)


src/condition/cooldown.rs
```diff
@@ -0,0 +1,233 @@
...
+/// Returns [`ActionState::None`] when the action timer is not finished or already actuated last frame.
+#[derive(Component, Reflect, Debug, Clone)]
+pub struct Cooldown {
+    /// Trigger threshold.
+    pub actuation: f32,
+    
+    actuated: bool,
+
+    /// The type of time used to advance the timer.
+    pub time_kind: TimeKind,
```
Shatur, line 18: Could you swap places these 2? So public fields come first :slightly_smiling_face:

src/condition/cooldown.rs
```diff
@@ -0,0 +1,233 @@
...
+
+    timer: Timer,
+}
+
+impl Cooldown {
+    #[must_use]
```
Shatur, line 24:
```suggestion
     /// Creates a new instance with the given cooldown time in seconds.
    #[must_use]
```

Shatur (changes requested):
Thank you! Looks very useful. Left a couple comments.

Could you also add this condition to `all_conditions` example?

src/condition/cooldown.rs
```diff
@@ -0,0 +1,233 @@
+use core::time::Duration;
+use bevy::prelude::*;
+
+use super::DEFAULT_ACTUATION;
+use crate::prelude::*;
+
+/// Returns [`ActionState::Fired`] when input is being pressed after a set duration passed or for the first time.
+///
+/// Returns [`ActionState::None`] when the action timer is not finished or already actuated last frame.
+#[derive(Component, Reflect, Debug, Clone)]
```
Shatur, line 10: We need to register it as reflected in the app.

examples/all_conditions.rs
```diff
@@ -2,6 +2,7 @@
 //! Press keys from the number row on the keyboard to trigger actions and observe the output in console.
 
 use bevy::{ecs::spawn::SpawnWith, log::LogPlugin, prelude::*};
+use bevy_enhanced_input::condition::cooldown::Cooldown;
```
Shatur, line 5: Ah, forgot to ask: add the type to `prelude`, so you won't need to import it here and in `src/lib.rs`.

examples/all_conditions.rs
```diff
@@ -66,6 +67,12 @@ fn spawn(mut commands: Commands) {
                 Tap::new(0.5),
                 bindings![TestTap::KEY],
             ));
+            context.spawn((
+                Action::<TestCooldown>::new(),
+                Cooldown::new(1.0),
+                bindings![TestCooldown::KEY],
+            ));
+
```
Shatur, line 75: I'd suggest to either put it last, to shift the assigned keys. So the ordering of 1,2,3...-,= will be consistent with action definition and spawning. It's just easier to navigate this way :slightly_smiling_face:

Shatur (changes requested):
Looks good, a few tiny suggestions :slightly_smiling_face:

idangur:
Just added `only_when_released` and checked it in my local implementation on the game and it looks fine

examples/all_conditions.rs
```diff
@@ -67,6 +67,7 @@ fn spawn(mut commands: Commands) {
                 bindings![TestTap::KEY],
             ));
 
+
```
Shatur, line 70:
```suggestion
```

Shatur (changes requested):
Great!
There are 2 unresolved comments left, please take a look :)
Also adding one more to remove an extra newline.

And I think we need a test for `only_when_released`.

I can made these change myself as well if you are busy, just let me know.

idangur:
@Shatur yeah I would be really glad if you can merge those as I am on vacation today, thanks!!

Shatur (approved):

idangur:
Looks great, just one question about the new logic, does it mean the timer won't tick when the button is held down?
And if I understand correctly it will tick all of the delta time that passed once it stop's actuating right?

Shatur:
Correct!
I didn't make it configurable because users can use conditions like `Press` or `Release` to make it fire for only one frame if necessary.

ekwoka:
I was just thinking yesterday how I wanted this and was going to try to make it.

PR #181, opened 2025-08-20
Add `Binding::AnyKey`

Tsudico:
This PR adds `Binding::AnyDigital` that is valid for any keyboard, mouse button, or gamepad button input. I have also included an example of how it can be used to determine player assignments for gamepads and in general.

While it respects bindings that come before it, it currently does not consume the input it is valid for so in the documentation there are specifications for its use. I'm seeking feedback so that we can determine whether a viable solution exists to properly consume the input so it can be used without issue and whether it should be added to BEI at all.


src/binding.rs
```diff
@@ -75,6 +75,13 @@ pub enum Binding {
     ///
     /// Useful for expressing empty bindings in [presets](crate::preset).
     None,
+    /// Any digital input, captured as [`ActionValue::Bool`]. If used with a context
+    /// that selects an input device, such as [`GamepadDevice::Single`], it will only
+    /// capture inputs from that device.
+    ///
+    /// This should only be used in the lowest priority context, after all other
+    /// actions, and there should only be one binding of this type active at a time.
+    AnyDigital,
```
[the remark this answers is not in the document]
Tsudico: I'm not opposed to changing it to `AnyKey` especially if it lines up with UE. I'll wait for Alice's thoughts though.

src/context/input_reader.rs
```diff
@@ -345,7 +427,7 @@ impl IgnoredInputs {
 
                 self.gamepad_axes.insert(input);
             }
-            Binding::None => (),
+            Binding::None | Binding::AnyDigital => (),
```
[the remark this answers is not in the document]
Tsudico: I've already mentioned how it might be able to be used for defined sequences or chains of actions, such as for fighting combos, using the `BlockBy` condition if sequences are ever considered within the scope of BEI. I'm not sure if it would be useful for quick time events or other dynamic input sequences. Those likely would look at each input and have external logic for whether that input was correct at the time.

Similar to the above, another possibility is with tutorials or guided events where you don't want to accept just the right input but also possibly the wrong input to give additional information or assistance to players when that occurs. In that case, you can put the proper action first which would be consumed and then have this binding afterward to accept any "wrong" input. I created something like that in my example. It is set to clear gamepads using the 'C' key but to accept any other input to change the color. In order to accomplish that the `ClearGamepads` action is first to catch the 'C' key and then the `Binding::AnyDigital` is used to change colors for any other key/button input.

So for the use cases I thought of, `Binding::AnyDigital` would need to have the lowest priority, preferably being the last to evaluate but consuming all inputs after it in case it wasn't. For "Press any key to continue" it seems like having a context with only the single binding by itself could be done to make it so the priority wouldn't matter.

In all the use cases we listed though, it seems that consuming all inputs when the binding activates wouldn't prevent the cases from being possible. So this might be the solution we can go with unless someone presents a use case that would require a binding to activate *after* `Binding::AnyDigital` is accepted.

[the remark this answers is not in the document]
Shatur: > chains of actions, such as for fighting combos, using the BlockBy condition

In fighting combos, the combo is only broken if you press another action, right? I think if you press a key that isn't bound to any action, it should be fine, isn’t it? So for sequences I'd expect users to block all other actions except the next one (not any input).

> Similar to the above, another possibility is with tutorials

This might be useful, agree.

> In all the use cases we listed though, it seems that consuming all inputs when the binding activates wouldn't prevent the cases from being possible.

With what I suggested, where the context priority equal to the number of all keyboard modifiers (4), it will evaluate first and block all other inputs, which breaks the mentioned use cases. Unless you disable input consumption for it. Would it be a possible workaround?
Without this, the use case I mention won't work properly because actions with modifier, like Ctrl + C, will be evaluated first.

If you think that the approach I suggested + workaround is not flexible enough, we can introduce a component that overrides action priority. By default it's the number of modifiers. But if this component is present - use the user-defined value. It should be a very simple change. This way we would by default put `Binding::AnyDigital`, but users can override it.

src/context/input_reader.rs
```diff
@@ -153,6 +153,91 @@ impl InputReader<'_, '_> {
                 value.into()
             }
             Binding::None => false.into(),
+            Binding::AnyDigital => {
```
Shatur, line 156: `get_pressed` already returns pressed keys, so I think we don't need to call things like `self.keys.any_pressed([*key])`.

I think we can simplify all this logic. Here is what I'd do:
1. Create iterators using `get_pressed` for keys, mouse_buttons, etc., with mapping their values into `Binding`. Use a separate variable for each iterator.
2. Chain all iterators, filter out all ignored values and call `next().is_some()`.

src/context/input_reader.rs
```diff
@@ -153,6 +153,59 @@ impl InputReader<'_, '_> {
...
+                if self.ignored(Binding::AnyDigital) {
+                    return false.into();
+                }
+                let value = match *self.gamepad_device {
+                    GamepadDevice::Single(entity) => {
```
Shatur, line 161: With `GamepadDevice::Single` I would still expect for other sources to be checked.

src/context/input_reader.rs
```diff
@@ -153,6 +153,59 @@ impl InputReader<'_, '_> {
...
+                            )
+                            .chain(
+                                self.mouse_buttons
+                                    .get_pressed()
+                                    .filter(|_| self.action_sources.mouse_buttons)
+                                    .map(|button| Binding::MouseButton {
+                                        button: *button,
+                                        mod_keys: ModKeys::empty(),
+                                    }),
+                            )
```
Shatur, line 202: Maybe extract into separate variables each iterator and then `chain` them? Would read better as for my taste.

Shatur (changes requested):

src/binding.rs
```diff
@@ -75,6 +75,16 @@ pub enum Binding {
     ///
     /// Useful for expressing empty bindings in [presets](crate::preset).
     None,
+    /// Any key, mouse button, or gamepad button, captured as [`ActionValue::Bool`].
+    /// If used with a context that selects an input device, such as
```
Shatur, line 79: Let's put a newline here. Usually in docs you have a short description and then details in additional paragraphs.
```suggestion
    /// Any key, mouse button, or gamepad button, captured as [`ActionValue::Bool`].
    ///
    /// If used with a context that selects an input device, such as
```

src/binding.rs
```diff
@@ -75,6 +75,16 @@ pub enum Binding {
...
+    /// [`GamepadDevice::Single`], it will only capture inputs from that device.
+    ///
+    /// When activated by the indicated inputs this binding will consume all remaining
+    /// inputs. Any actions at a lower priority, whether in the same context or lower
+    /// priority contexts, will not activate. If you want this binding to activate
+    /// prior to bindings that include modifier keys then it needs to be placed in a
+    /// higher priority context.
```
Shatur, line 86: How about to rewrite it like this? Because other bindings, such as mouse movement, may still trigger other actions.
```suggestion
    /// If [`ActionSettings::consume_input`] is set, this binding consumes all button
    /// inputs, not just the one that activated it. To have an action with this binding
    /// evaluated first, place it in a higher-priority context.
```

src/context/input_reader.rs
Tsudico: I was unable to get the iterators into separate variables, possibly due to lack of experience with Rust. Either the iterator was no longer recognized as an iterator but something like "struct &FlatMap<..." or the iterators did not have the same type anymore so could not be chained.

I understand the concern for clarity though. I'll have to research more to see if I can find a solution.

Shatur: Yes, each iterator will have a different type. So you can't use things like `if`, you have to apply `filter` like you did.
I think if you just write something like

```rust
        let keys = self
            .keys
            .get_pressed()
            .filter(|_| self.action_sources.keyboard)
            .map(|key| Binding::Keyboard {
                key: *key,
                mod_keys: ModKeys::empty(),
            });
```

then it will work.

src/context/input_reader.rs
Shatur: `GamepadDevice::Single` only limits the gamepad input to specific device. All other input sources still used.
You can take a look at the `local_multiplayer` example. It doesn't use `AnyDigital`, but that's how other bindings work and I think we need to make `AnyDigital` consistent.

Checking the gamepad input would require another solution, but in this case I'd just use `bevy_input` instead of spawning a context for each connected gamepad.

src/context/input_reader.rs
Tsudico: That explains it. When I tried to separate them I also removed the filter because I thought I could use `if` on the action_sources. I will update this when I make a new commit.

src/context/input_reader.rs
```diff
@@ -153,6 +153,58 @@ impl InputReader<'_, '_> {
...
+                        mod_keys: ModKeys::empty(),
+                    });
+                let mouse_buttons = self
+                    .mouse_buttons
+                    .get_pressed()
+                    .filter(|_| self.action_sources.mouse_buttons)
+                    .map(|button| Binding::MouseButton {
+                        button: *button,
+                        mod_keys: ModKeys::empty(),
+                    });
```
Shatur, line 175: I think you can use this here and in other places.
```suggestion
                    .map(Into::into);
```

Shatur (changes requested):
I like the progress so far!

Left a small suggestion. Tomorrow I'll think if we can simplify the gamepad selection logic and I think it will be ready to merge.

Left a few small suggestions, check unresolved dialogs.

src/context/input_reader.rs
Tsudico: Trying to replace it with that gives me an error:
```
error[E0277]: the trait bound `binding::Binding: From<&bevy::prelude::KeyCode>` is not satisfied
160 |                   let keyboard = self
    |  ________________________________^
161 | |                     .keys
162 | |                     .get_pressed()
163 | |                     .filter(|_| self.action_sources.keyboard)
164 | |                     .map(Into::into);
    | |____________________________________^ the trait `From<&bevy::prelude::KeyCode>` is not implemented for `binding::Binding`
```

src/binding.rs
```diff
@@ -184,15 +197,19 @@ impl<I: Into<Binding>> InputModKeys for I {
...
             Binding::MouseButton { button, .. } => Binding::MouseButton { button, mod_keys },
             Binding::MouseMotion { .. } => Binding::MouseMotion { mod_keys },
             Binding::MouseWheel { .. } => Binding::MouseWheel { mod_keys },
-            Binding::GamepadButton { .. } | Binding::GamepadAxis { .. } | Binding::None => {
-                panic!("keyboard modifiers can be applied only to mouse and keyboard")
+            Binding::GamepadButton { .. }
+            | Binding::GamepadAxis { .. }
+            | Binding::None
+            | Binding::AnyKey => {
+                panic!("keyboard modifiers can be applied only to mouse and keyboard bindings")
```
alice-i-cecile, line 212: Non-blocking: this should not panic. It is fully recoverable: just log a warning and then retrun `Self`.

alice-i-cecile (approved):
Good addition! This is a handy helper.

PR #188, opened 2025-08-28
Make ActionTime pub

cBournhonesque:
For input prediction purposes, I need this method to be public.

The reason is that:
- on client I am on tick 9
- I have received the inputs from another client for tick 5
- To correctly predict the actions of the other client, currently I act as if the remote client keeps triggering the same actions. So I predict that on ticks 6, 7, 8, 9 the ActionValue, ActionState, ActionEvents are the same. However I need to update the ActionTime correctly.

Let me know if this makes sense. It's a pretty ad-hoc approach, ideally the ActionState/ActionEvents would also be updated correctly (for example I wouldn't want to predict that a rare event is emitted every tick)

### DIFF
--- a/src/action.rs
+++ b/src/action.rs
@@ -259,7 +259,7 @@ pub struct ActionTime {
 }
 
 impl ActionTime {
-    pub(crate) fn update(&mut self, delta_secs: f32, state: ActionState) {
+    pub fn update(&mut self, delta_secs: f32, state: ActionState) {
         match state {
             ActionState::None => {
                 self.elapsed_secs = 0.0;
--- a/src/action.rs
+++ b/src/action.rs
@@ -259,6 +259,8 @@ pub struct ActionTime {
 }
 
 impl ActionTime {
+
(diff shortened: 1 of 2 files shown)


Shatur:
Why do you need to call `ActionTime::update` manually? I thought you were just running the `EnhancedInputSet::Update` system during rollback.

src/action.rs
```diff
@@ -259,7 +259,9 @@ pub struct ActionTime {
 }
 
 impl ActionTime {
-    pub(crate) fn update(&mut self, delta_secs: f32, state: ActionState) {
+
```
Shatur, line 262:
```suggestion
```

Shatur (changes requested):
Ah, got it, makes total sense.

> For ActionEvent, I should also probably predict it from the last 2 ActionState.

Yeah, I'd do the same.

Left a small nitpick.

Shatur (approved):

PR #189, opened 2025-08-29
Add ExternallyMocked

cBournhonesque:
In niche situations (networking), it is useful to be able to opt-out of running `BEI::Update` for some Action entities.

For example if we want to predict the actions of remote clients, we receive from the Server regular messages that give us information about the ActionState for these other clients.

For these remote clients, we don't want to run BEI::Update but instead we want to have greater control over how they are updated:
- their state could be pulled from an input buffer if we have that available
- or we could decay their action slightly if we don't have any ticks available

An alternative could be to set an ActionMock on these remote entities, but that's not efficient because it would require running BEI::Update during rollbacks so that the ActionMock can update the ActionTime/ActionEvents components.
Therefore we provide an escape hatch called `ExternallyMocked` to indicate that the client has control over how the components of that Action are updated.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -180,7 +180,6 @@ impl ScheduleContexts {
             QueryParamBuilder::new(|builder| {
                 builder
                     .data::<Option<&GamepadDevice>>()
-                    .without::<ExternallyMocked>()
                     .optional(|builder| {
                         for &id in &self.activity_ids {
                             builder.mut_id(id);
@@ -194,7 +193,7 @@ impl ScheduleContexts {
             ParamBuilder,
             ParamBuilder,
             QueryParamBuilder::new(|builder| {
-                builder.without::<ExternallyMocked>().optional(|builder| {
+                builder.optional(|builder| {
                     for &id in &**conditions {
                         builder.mut_id(id);
                     }
--- a/src/context.rs
+++ b/src/context.rs
@@ -194,9 +194,7 @@ impl ScheduleContexts {
             ParamBuilder,
             ParamBuilder,
             QueryParamBuilder::new(|builder| {
-                builder
-                    .without::<ExternallyMocked>()
-                    .optional(|builder| {
+                builder.without::<ExternallyMocked>().optional(|builder| {
                     for &id in &**conditions {
                         builder.mut_id(id);
                     }
--- a/src/context.rs
+++ b/src/context.rs
@@ -340,11 +340,11 @@ pub(crate) fn reset_action<C: Component>(
     }
 }
 
-/// Marker component that can be used to exclude some [`Action`] entities from the `update` system.
+/// Marks an [`Action<C>`] as manually mocked, skipping the [`EnhancedInputSet::Update`] logic for it.
 ///
-/// This is useful if the user wants greater control over how the [`Action`]'s components are updated.
+/// This allows modifying any action data without its values being overridden during evaluation.
 ///
-/// This is an advanced feature that should be used with caution.
+/// Takes precedence over [`ActionMock`], which drives specific [`ActionValue`] and [`ActionState`] during evaluation.
 #[derive(Component)]
 pub struct ExternallyMocked;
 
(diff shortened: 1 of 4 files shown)


src/context.rs
```diff
@@ -340,22 +340,33 @@ pub(crate) fn reset_action<C: Component>(
     }
 }
 
+/// Marker component that can be used to exclude some [`Action`] entities from the `update` system.
+///
+/// This is useful if the user wants greater control over how the [`Action`]'s components are updated.
+///
+/// This is an advanced feature that should be used with caution.
```
Shatur, line 347:
```suggestion
/// Marks an [`Action<C>`] as manually mocked, skipping the [`EnhancedInputSet::Update`] logic for it.
///
/// This allows modifying any action data without its values being overridden during evaluation.
///
/// Takes precedence over [`ActionMock`], which drives specific [`ActionValue`] and [`ActionState`] during evaluation.
```

Shatur (changes requested):
I like the new name, but docs need a little bit of work.

Could you also add something like "See also ExternallyMocked to manually control the action data" to `ActionMock` docs?

Shatur:
Could you also add a small test to `tests/mocking.rs`?

src/context.rs
```diff
@@ -180,6 +180,7 @@ impl ScheduleContexts {
             QueryParamBuilder::new(|builder| {
                 builder
                     .data::<Option<&GamepadDevice>>()
+                    .without::<ExternallyMocked>()
```
Shatur, line 183: This query is for contexts.

tests/mocking.rs
```diff
@@ -132,6 +133,37 @@ fn manual() {
...
+
+    app.update();
+
+    let mut actions = app
+        .world_mut()
+        .query::<(&Action<Test>, &ActionState, &ActionEvents)>();
+
+    // the action data should not be updated because of ExternallyMocked
+    let (&action, &state, &events) = actions.single(app.world()).unwrap();
+    assert!(!*action);
```
Shatur, line 162:
```suggestion
    let (&action, &state, &events) = actions.single(app.world()).unwrap();
    assert!(!*action, "action shouldn't be updated because it marked as mocked externally");
```

src/context.rs
```diff
@@ -193,7 +194,7 @@ impl ScheduleContexts {
             ParamBuilder,
             ParamBuilder,
             QueryParamBuilder::new(|builder| {
-                builder.optional(|builder| {
+                builder.without::<ExternallyMocked>().optional(|builder| {
```
Shatur, line 197: Why disable it for the conditions and modifiers query? If the action is skipped, it won't be used anyway.

Shatur (changes requested):

src/context.rs
cBournhonesque: oops, you're right

src/context.rs
cBournhonesque: you're right

tests/mocking.rs
```diff
@@ -1,6 +1,7 @@
 use core::time::Duration;
 
 use bevy::{prelude::*, time::TimeUpdateStrategy};
+use bevy_enhanced_input::context::ExternallyMocked;
```
Shatur, line 4: One last suggestion! Let's add `ExternallyMocked` to the `prelude`.

Shatur: Agree!

cBournhonesque:
> You mentioned you getting warnings for actions marked as externally mocked? Let's remove the warning and I think the PR is ready.

I don't know how to revemo the warning efficiently though; by doing `Has<ExternallyMocked>` instead of `Without<ExternallyMocked>` in the query? But then we lose the performance benefits.

Shatur (approved):
I meant to remove it literally, which is exactly what you did :slightly_smiling_face:

The PR looks good and the use case is clear.

cBournhonesque:
> Ah, could you remove the unused import to make CI pass?

yep!

Shatur:
Ah, the PR was merged because I enabled auto-merge and approved it. @alice-i-cecile let us know here if you are fine with this change. Sorry for the mess.

PR #209, opened 2025-09-17
Starting migration to Bevy 0.17

Lenchog:
There are still issues with spawning as `move_as_ptr` is broken upstream, apart from that the errors and warnings are gone.

PR #210, opened 2025-09-24
Migrate to Bevy 0.17.0-rc

janhohenheim:
Adopted #209


.github/workflows/main.yml
```diff
@@ -84,7 +84,7 @@ jobs:
         uses: Swatinem/rust-cache@v2
 
       - name: Check compilation
-        run: cargo check --target thumbv6m-none-eabi --features bevy/critical-section
+        run: cargo check --target thumbv6m-none-eabi --features bevy/critical-section --lib
```
janhohenheim, line 87: so we don't pull in dev deps

README.md
```diff
@@ -29,6 +29,7 @@ Have any questions? Feel free to ask in the dedicated [`bevy_enhanced_input` cha
 
 | bevy        | bevy_enhanced_input |
 | ----------- | ------------------- |
+| 0.17.0-rc      | 0.19.0-rc           |
```
janhohenheim, line 32: not bothering with formatting since the table would need to be formatted back when the RC is over

.github/workflows/main.yml
Shatur: Ah, got it, agree!

PR #211, opened 2025-09-24
Update to Bevy 0.17.0-rc.2

Shatur:
Will be merged once Bevy releases.


Shatur:
I planning to merge the PR today, just want to get #216 in the release.

But you don't have to wait, RC versions are compatible with releases. So you can use the RC version of this crate with Bevy 0.17 (non-RC).

PR #224, opened 2025-10-14
Make crate documentation more approachable

alice-i-cecile:
## Problem

I've done an editing pass on the crate documentation! While this contained a lot of excellent information and helpful links, I think that the existing structure was suboptimal, and the foundational concepts were not given the time it deserves.

## Solution

- general purpose editing and writing pass
- move as much of the content as we can to linked submodules
- spend more time really explaining the core context/actions/bindings structure
- be more explicit about how the pull-based API is used

## Future work

- revise moved module docs, especially for Context
- create and link to a dedicated minimal example
- remove the Input and UI section, probably by integrating with bevy_input_focus!

### DIFF
--- a/src/preset.rs
+++ b/src/preset.rs
@@ -72,8 +72,6 @@ world.spawn((
 ));
 ```
 
-You can also assign custom bindings or attach additional modifiers, see the [preset] module for more details.
-
 # Implementation details
 
 Each of the preset types defined in this module generates a list of binding entities with the appropriate components
--- a/src/modifier.rs
+++ b/src/modifier.rs
@@ -6,7 +6,7 @@ Action values are stored in two forms:
 
 During [`EnhancedInputSet::Update`], input is read for each [`Binding`] as an [`ActionValue`], with the variant depending
 on the input source. This value is then converted into the [`ActionValue`] on the associated action entity. For example,
-key inputs are captured as [`bool`], but if the action's output type is [`Vec2`], the value will be assigned to the X axis
+key inputs are captured as [`bool`], but if the action's output type is [`Vec2`](bevy::math::Vec2), the value will be assigned to the X axis
 as `0.0` or `1.0`. See [`Binding`] for details on how each source is captured, and [`ActionValue::convert`] for how values
 are transformed.
 


src/lib.rs
```diff
@@ -1,58 +1,68 @@
...
+#[action_output(bool)]
+struct Jump;
+
+#[derive(InputAction)]
+#[action_output(bool)]
+struct Fire;
+
+let mut app = App::new();
+app.add_plugins(EnhancedInputPlugin);
 app.add_input_context::<Player>();
```
Shatur: The example includes context registration, but the context section comes later. This feels more like an example for "Putting it all together" section. But then it will look like a lot of text before actually showing the code :thinking: 

What do you think about this?
- Move bindings section first and put a very short example where bindings are defined with the mentioned macro. Nothing else, just bindings.
- For actions we showcase how actions are defined for these bindings. No context yet.
- For the context section we wrap it into a context and showcase the spawn and registration.

src/lib.rs
```diff
@@ -1,58 +1,68 @@
...
 
-let mut app = App::new();
-app.add_plugins((MinimalPlugins, EnhancedInputPlugin));
-```
+Each action represents a different in-game behavior.
+To create a new action, you will need to define a new struct and implement the [`InputAction`] trait for it,
+typically using the provided [`InputAction`] derive macro.
 
-## Core Concepts
+The action's output type is defined by the [`InputAction::Output`] associated type,
```
Shatur: Maybe mention that actions consume the binding by default? This important for context ordering.

src/lib.rs
```diff
@@ -67,202 +77,130 @@ world.spawn((
...
-on the input source. This value is then converted into the [`ActionValue`] on the associated action entity. For example,
-key inputs are captured as [`bool`], but if the action's output type is [`Vec2`], the value will be assigned to the X axis
-as `0.0` or `1.0`. See [`Binding`] for details on how each source is captured, and [`ActionValue::convert`] for how values
-are transformed.
+Contexts can be [layered](crate::context::ContextPriority), causing one context to rely on another's presence, and any number of them can be active at the same time.
+To register a component as an input context, you need to call [`InputContextAppExt::add_input_context`]. By default, contexts are
+evaluated during [`PreUpdate`], but you can customize this by using [`InputContextAppExt::add_input_context_to`] instead.
 
-Then, during [`EnhancedInputSet::Apply`], the value from [`ActionValue`] is written into [`Action<C>`].
+Context actions will be evaluated in the schedule associated at context registration. Contexts registered in the same
```
Shatur: Looks like this paragraph duplicates some of the information from the previous two.

src/lib.rs
```diff
@@ -67,202 +77,130 @@ world.spawn((
...
+Most users find the push-style API more ergonomic and easier to reason about,
+but the pull-style API can allow for more complex checks and interactions between the state of multiple actions.
 
-## Reacting on actions
+Ultimately, the choice between these two approaches depends on your specific use case and preferences,
+with performance playing a relatively minor role unless you have a very large number of acting entities or if you have a complex logic for your action reaction.
 
-Up to this point, we've only defined actions and contexts but haven't reacted to them yet.
-We provide both push-style (via observers) and pull-style (by checking components) APIs.
+## Push-style
```
Shatur:
```suggestion
### Push-style
```

src/lib.rs
```diff
@@ -67,202 +77,130 @@ world.spawn((
...
-## Input conditions
+While we can bind actions directly to input sources like buttons, further customization or preprocessing is often needed.
+Not all inputs are "buttonlike": we may want to only trigger an action when a button is held for a certain duration,
+or when a joystick is moved beyond a certain threshold.
 
-Instead of hardcoded states like "pressed" or "released", all actions use an abstract [`ActionState`] component
-(which is a required component of [`Action<C>`]). Its meaning depends on the assigned [input conditions](crate::condition),
-which determine when the action is triggered. This allows you to define flexible behaviors, such as "hold for 1 second".
+There are two main ways to achieve this: using input conditions and input modifiers:
+- Input conditions define when an action is considered to be triggered based on the state of its bindings.
```
Shatur: We probably need to mention that without any input condition it works like `Down` without any actuation threshold.

Shatur (changes requested):
I think this might work, yes.
Left a few suggestions.

My only worry now is that we now have quite a lot of text where we just explain terms before we showcase the code :thinking: But not sure what to suggest about it.

src/preset.rs
```diff
@@ -1,13 +1,94 @@
...
+struct Movement;
+```
+
+However, this is quite onerous! It would be inconvenient to bind WASD keys and analog sticks manually, like in the example above,
+every time. You can use [`Cardinal`](crate::Cardinal) and [`Axial`](crate::Axial) presets to simplify the example above.
+
+```
+use bevy::prelude::*;
+use bevy_enhanced_input::prelude::*;
+let mut world = World::new();
```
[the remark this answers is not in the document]
alice-i-cecile: This is deliberately unhidden. Hiding this sort of thing is really confusing for newer users. Agreed on whitespace though!

src/lib.rs
```diff
@@ -67,202 +77,130 @@ world.spawn((
...
 
-Actions also have [`ActionSettings`] component that customizes their behavior.
+Contexts define when actions are evaluated. They are associated with action entities via the [`Actions<C>`] relationship,
+where `C` is your context type, and can be quickly bound to them using the [`actions!`] macro.
 
-## Input modifiers
+Inside of this crate, contexts are stored using regular components, commonly on an entity for which the input is associated (player, button, dialog, etc.).
+Depending on your type of game, you may have a single global context
+or multiple contexts for different gameplay states. For games with multiple entities driven by a single context it's
+common to create a "controller" entity which applies the input to the desired entity.
```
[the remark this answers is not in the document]
alice-i-cecile: I think that this is really important foundational knowledge about "what is a context".

src/lib.rs
```diff
@@ -1,268 +1,211 @@
...
-use bevy::prelude::*;
-use bevy_enhanced_input::prelude::*;
+When actions are triggered, the "consume" the input values from their bindings.
+This means that if multiple actions are bound to the same input source (e.g., the same key),
+the action that is evaluated first will take precedence, and the others will not receive the input value.
 
-# let mut app = App::new();
-# app.add_plugins(EnhancedInputPlugin);
-app.add_input_context::<Player>();
+Inside of this crate, actions are stored as entities with the [`Action<A>`] component, where `A` is your [`InputAction`] type.
```
Shatur: Maybe simplify?
```suggestion
Actions are stored as entities with the [`Action<A>`] component, where `A` is your [`InputAction`] type.
```

src/lib.rs
```diff
@@ -1,268 +1,211 @@
...
+or multiple contexts for different gameplay states. For games with multiple entities driven by a single context it's
+common to create a "controller" entity which applies the input to the desired entity.
+
+Inside of this crate, contexts are stored using regular components, commonly on an entity for which the input is associated (player, button, dialog, etc.).
+
+Contexts can be activated or deactivated using the [`ContextActivity`] component.
+By default, contexts are active when the component is present.
+When active, all actions associated with the context are evaluated.
+
+Contexts can be [layered](crate::context::ContextPriority), defining a priority order, and any number of them can be active at the same time.
```
Shatur: We previously mentioned that multiple contexts can be active at the same time. Maybe write it like this?
```suggestion
By default contexts evaluated in their spawn order, but you can also control this with [`ContextPriority`].
```

src/lib.rs
```diff
@@ -1,268 +1,211 @@
...
+or when a joystick is moved beyond a certain threshold.
+
+There are two main ways to achieve this: using input conditions and input modifiers:
+
+- Input conditions define when an action is considered to be triggered based on the state of its bindings.
+  For example, the [`Hold`] condition can be used to trigger an action only when a button is held down for a specified duration.
+- Input modifiers transform the raw input values from bindings before they are processed by the action.
+  For example, the [`DeadZone`] modifier can be used to ignore small movements of a joystick.
+
+When no input conditions are attached to an action or its bindings,
```
Shatur: Maybe put merge this paragraph with the input condition list item above?

src/lib.rs
```diff
@@ -1,268 +1,211 @@
...
-To register a component as an input context, you need to call [`InputContextAppExt::add_input_context`]. By default, contexts are
-evaluated during [`PreUpdate`], but you can customize this by using [`InputContextAppExt::add_input_context_to`] instead.
+In short, actions are mapped to inputs via bindings, and contexts control which actions are active.
 
-Actions are represented by entities with the [`Action<A>`] component, where `A` is a user-defined marker that implements the
-[`InputAction`] trait, which defines [`InputAction::Output`] type - the value the action produces. It could be [`bool`], [`f32`],
-[`Vec2`] or [`Vec3`]. Actions associated with contexts via [`ActionOf`] relationship. We provide the [`actions!`] macro, which is
-similar to [`related!`], but for actions. The relationship is generic over `C` because a single entity can have multiple associated
-contexts.
+## Actions
```
Shatur: Have you considered swapping actions and bindings to go from the bottom of the hierarchy? Or do you think this reads better when actions come first? Just asking :slightly_smiling_face:

alice-i-cecile: Bindings only make sense as a link between inputs and actions. If users don't know what actions are, they'll be lost!

PR #225, opened 2025-10-15
Initial delay

newclarityex:
Adds initial delay on Pulse, as well as reset_on_change.
Initial delay requires the action to be actuated for a certain amount of time before pulsing (DAS)
Reset on change will reset the above duration when the actuated value changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - `Combo` input condition.
+- `Pulse::with_initial_delay` and `Pulse::initial_delay` to add special delay before the first repeat.
 
 ### Changed
 


Shatur (changes requested):
I think it would be better to replace `trigger_on_start` with enum. Because right now `initial_delay` does nothing if `trigger_on_start` isn't set.

`reset_on_change` is weird. What is your use case?

newclarityex:
In what way does initial_delay do nothing if trigger_on_start isn't set? The expected behavior should be if trigger on start + delay is set, it triggers right away then triggers after the delay. If trigger on start isn't set, it should wait until the delay before triggering. Am I misunderstanding trigger_on_start?

Also reset_on_change was for a usecase where changing directions would allow the delay to reset for more precise movement, but after some testing I don't think I need it.

newclarityex:
> Looks like you're ticking both timers at the same time. In your test, the initial time is divisible by the regular delay. But if you write it like this, it fails, even though I'd expect it to pass: [test.patch](https://github.com/user-attachments/files/22954749/test.patch) (apply with `git patch <path to file>`).

Not sure what you mean by this, and git patch didn't work for me. Could you post what the test looks like?

> I think we should tick the initial timer first, and then tick the regular one. Also, let's use `Timer` instead of `Duration`.

Wouldn't a timer technically allow the user to write a Repeating timer, even if it doesn't make sense here? What's the benefit over a `Duration`?

Shatur:
Adjusted the behavior myself. Should be ready to merge.

alice-i-cecile (approved):

PR #229, opened 2025-10-20
Add `Combo` input condition

Shatur:


src/condition/combo.rs
```diff
@@ -0,0 +1,552 @@
...
+    /// If a cancel action matches the action from the current step, it will be ignored.
+    pub cancel_actions: Vec<CancelAction>,
+
+    /// The type of time used to advance the timer.
+    pub time_kind: TimeKind,
+
+    /// Index of the current step in the combo.
+    step_index: usize,
+
+    timer: Timer,
```
alice-i-cecile: A more descriptive name here would be nice.

alice-i-cecile: Also doc comments!

alice-i-cecile:
I really like this: this is very common in some games and gnarly to implement.

src/condition/combo.rs
```diff
@@ -0,0 +1,552 @@
...
+
+use bevy::prelude::*;
+use log::warn;
+
+use crate::prelude::*;
+
+/**
+Sequence of actions that needs to be triggered in specific order.
+
+The condition activates for one frame before resetting the combo progress.
```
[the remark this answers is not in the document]
Shatur: Sorry, it's just a bad wording :sweat_smile:
I rephrased it, should be clear now.

src/condition/combo.rs
Shatur: I added doc comment, but kept the name.
I'm just not sure how else to call it. But I think with the doc comment the `timer` name works :thinking:

PR #230, opened 2025-10-20
Module level docs for action.rs

alice-i-cecile:
More docs, more good! Some of this is a repeat from the crate docs, but that's fine: these are core concepts and it's important to provide adequate breadcrumbs.

### DIFF
--- a/src/action.rs
+++ b/src/action.rs
@@ -43,7 +43,7 @@
 //! using the [`ActionMock`] component or by directly modifying various components before [`EnhancedInputSystems`] are run.
 //!
 //! This is useful for simulating input during cutscenes,
-//! testing, AI-controlled players, tool-assisted speedruns or other scenarios where you want to control the action state directly.
+//! testing, networked replication, AI-controlled players, game replays, or other scenarios where you want to control the action state directly.
 
 pub mod events;
 pub mod fns;


src/action.rs
```diff
@@ -1,3 +1,58 @@
...
+//! The exact semantic meaning of these events (and [`ActionState`], discussed below),
+//! depends on the [conditions](crate::condition) defined on the action
+//! or the input bindings for that action.
+//!
+//! ## Checking the state of actions
+//!
+//! The current state of an action can be accessed using the [`ActionState`] component,
+//! which indicates whether the action is inactive, ongoing, or fired.
+//!
+//! The [`ActionValue`] component holds the current value of the action,
```
Shatur: `ActionValue` stores an untyped value. Users most likely want to access `Action<C>`.

src/action.rs
```diff
@@ -1,3 +1,58 @@
...
+//! Actions belong to [contexts](crate::context) that group related actions together,
+//! allowing you to enable and disable actions based on the current game state.
+//!
+//! They are spawned as entities with the [`Action<C>`] component, where `C` is the action type,
+//! and related to the context entity via the [`ActionOf<C>`] relationship.
+//!
+//! In turn, actions have input mappings defined by [bindings](crate::binding) entities,
+//! which are related to the action entity via the [`BindingOf`] relationship.
+//!
+//! ## Responding to actions
```
Shatur: Heading levels needs to be adjusted.
```suggestion
//! # Responding to actions
```

Shatur (changes requested):
I think we duplicate too much information here. I like the balanced approach you took for the `binding` module: just a general explanation what it is and links to the useful places. I'd suggest to do the same here. We can link to specific sections of the quick start guide via `../index.hml#section-name`.

But I left a few nitpics in case we decide to go with the duplication.

alice-i-cecile:
I've resolved the nits :)

Generally I'd prefer to have the bulk of the information here, and reduce duplication by cutting information from the more general library-level docs.  I think that "users who want to know more about how actions work" should / will be looking here, not at the broader docs.

That said, I think we can compromise. Discussions of "responding to actions" feels like a nice high-level concern, and definitely shouldn't be duplicated. I've chosen to leave that in the crate-level docs, and then linked back to it from the action docs.

src/action.rs
```diff
@@ -1,3 +1,45 @@
...
+//!
+//! # Responding to actions
+//!
+//! When an action is evaluated, it produces various [action events](events) that indicate
+//! changes in the action's state.
+//! See the section on [push-style action handling](../index.html#push-style-responding-to-action-events) in the library documentation for more details.
+//!
+//! Similarly, you can check the current state and value of an action at any time using the
+//! [`Action<C>`], [`ActionState`], [`ActionValue`] and [`ActionTime`] components.
+//! See the section on [pull-style action handling](../index.html#pull-style-polling-action-state) in the library documentation for more details.
```
Shatur: Same here.

src/action.rs
```diff
@@ -1,3 +1,45 @@
...
+//!
+//! The behavior of actions can also be modified via
+//! [modifiers](crate::modifier) that transform the action value during evaluation,
+//! or by using [input conditions](crate::condition) to control when actions are triggered.
+//!
+//! # Manually firing actions
+//!
+//! In addition to responding to user input, you can also manually set the state and value of actions
+//! using the [`ActionMock`] component. This is useful for simulating input during cutscenes,
+//! testing, AI-controlled players, tool-assisted speedruns or other scenarios where you want to control the action state directly.
```
Shatur: In addition to mocking, users can also directly modify components in a system before `EnhancedInputSystems`.
Maybe also worth mentioning?

And could you add a newline before `pub mod`? Just for consistency.

PR #240, opened 2025-10-29
Add with_mod_keys() to presets

### LINKED ISSUE #239: Add a helper to Cardinal to support mod keys
My original proposal follows, but Shatur suggests maybe a method on `Cardinal` instead.

A helper like this could be useful to add:

```rust
fn cardinal_with_mod_keys(
    cardinal: Cardinal<Binding, Binding, Binding, Binding>,
    mod_keys: ModKeys,
) -> Cardinal<Binding, Binding, Binding, Binding> {
    Cardinal {
        north: cardinal.north.with_mod_keys(mod_keys),
        east: cardinal.east.with_mod_keys(mod_keys),
        south: cardinal.south.with_mod_keys(mod_keys),
        west: cardinal.west.with_mod_keys(mod_keys),
    }
}
```

Example:

```rust
Bindings::spawn(cardinal_with_mod_keys(Cardinal::arrows(), ModKeys::CONTROL))
```

Not sure if there's a more ergonomic design than this being a helper function.

See [Discord](https://discord.com/channels/691052431525675048/1297361733886677036/1433064707505389711).

mgi388:
Fixes #239

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `Combo` input condition.
 - `Pulse::with_initial_delay` and `Pulse::initial_delay` to add special delay before the first repeat.
+- `with_mod_keys()` method to all preset types (`Cardinal`, `Bidirectional`, `Ordinal`, `Spatial`, `Axial`) to apply keyboard modifiers to all bindings in the preset.
 
 ### Changed
 


examples/basic_action_management.rs
```diff
@@ -176,6 +199,10 @@ struct FlyCam;
 #[action_output(Vec2)]
 struct Movement;
 
+#[derive(InputAction)]
+#[action_output(Vec2)]
+struct Sprint;
```
Shatur: I'm not sure if I would implement sprint this way :thinking: 
I'd probably make sprint a simple bool and just check its state in the movement observer.

Shatur (changes requested):
Maybe add this to all presets? :thinking:

examples/basic_action_management.rs
mgi388: I thought it might be a good way to demonstrate it, but I'd rather not include a convoluted example so I will just remove it.

alice-i-cecile (approved):
We could probably make these methods all const, but I'm not convinced that's super useful. LGTM.

Probably worth adding a quick line to the release notes?

mgi388:
It doesn't seem possible to make them const unless I am doing something wrong.

Shatur (approved):
The use case makes sense to me!

I'll draft a new patch release today after work.

PR #242, opened 2025-10-30
PoC implementing exact modifier key matching for input bindings

mgi388:
This PR is meant as a PoC and place to start a discussion about this, not necessarily as the proposed solution that should be merged. I chose to do a PR rather than an issue or discussion because I can include a runnable example which includes one potential fix.

## Problem

Bindings _without_ any modifier keys specified will incorrectly match when modifier keys are pressed by the user.

### Example scenario

(This is the specific scenario that matches my own use case).

Consider a game with two contexts:

- **Player context** (priority 0): `ToggleMute` bound to `Shift+M`.
- **Battle context** (priority 1): `ToggleMagicPanel` bound to `M`.

When in the battle context:

**Before this PR**

- Pressing `Shift+M` triggers **both** actions (noting that `ToggleMagicPanel` has `consume_input: false` even though the player's intent was to toggle mute. The player observes their magic panel open/close, and the audio is muted/unmuted.
- The binding `M` matches even though the `Shift` key was pressed.

**After this PR**

- Pressing `Shift+M` triggers **only** the toggle mute action.
- Pressing `M` (without Shift) triggers **only** the magic panel action.
- Bindings require **exact** modifier matches.

## Breaking change

This technically would be a **breaking behavioral change** because if a player/developer was arbitrarily holding down a modifier key while pressing an action's key, the action will no longer fire.

However, maybe it's also more of a bugfix: I defined my own `ToggleMagicPanel` action to be bound to `M`, _not_ `Shift+M`, or `Control+M`, or `Control+Shift+M`, or any other combination of modifiers—I just defined the action to be bound to `M` alone.

## Alternatives

I haven't really thought it through because I would rather focus on the actual problem right now, but at least one other solution here is probably to allow users to opt in or out of this behavior using some sort of `ModifierMatchBehavior` component/type.

## Updated example

I updated the `context_layering.rs` example to be more or less the same as my own use case. You can run it with and without the "fix" to see how the behavior changes.

### DIFF
--- a/examples/context_layering.rs
+++ b/examples/context_layering.rs
@@ -1,15 +1,41 @@
-//! Demonstrates the concept of context layering in input handling.
-//! One context can be applied on top of another, overriding some of the bindings.
+//! Demonstrates the concept of context layering in input handling. One context
+//! can be applied on top of another, overriding some of the bindings.
 //!
-//! The [`ContextPriority`] component is used to determine the order of contexts,
-//! with higher priority contexts taking precedence over lower priority ones.
-//! This influences the order in which actions are evaluated and inputs are consumed.
-//! See [`ActionSettings::consume_input`] for more details and control over this behavior.
+//! The [`ContextPriority`] component is used to determine the order of
+//! contexts, with higher priority contexts taking precedence over lower
+//! priority ones. This influences the order in which actions are evaluated and
+//! inputs are consumed. See [`ActionSettings::consume_input`] for more details
+//! and control over this behavior.
 //!
-//! In this example, we have a [`Player`] context that allows basic movement and jumping.
-//! When the player enters a vehicle, we add a [`Driving`] context on top of the [`Player`] context.
-//! The [`Driving`] context overrides the jump action with a brake action and adds actions for entering
-//! and exiting the vehicle.
+//! In this example, we have a [`Player`] context that allows basic movement,
+//! jumping, and muting audio. When the player enters a vehicle, we add a
+//! [`Driving`] context on top of the [`Player`] context. The [`Driving`]
+//! context overrides the jump action with a brake action and adds actions for
+//! entering and exiting the vehicle, plus a map display action.
+//!
+//! # Exact Modifier Key Matching
+//!
+//! This example also demonstrates exact modifier key matching:
+//! - The [`Player`] context has a [`ToggleMute`] action bound to `Shift+M`
+//! - The [`Driving`] context has a [`ToggleMap`] action bound to just `M` (no
+//!   modifiers)
+//! - When driving and pressing `Shift+M`, only the mute action fires (not the
+//!   map action)
+//! - When driving and pressing just `M`, only the map action fires
+//!
+//! This works because bindings require **exact** modifier matches:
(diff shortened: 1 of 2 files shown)


Shatur:
I see the problem, but this behavior is a bit unexpected. Example: a racing game with nitro bound to Shift and movement to WASD. Once you press Shift, the car will simply stop moving. We can make this configurable via `ActionSettings`, or explore other alternatives...

P.S. I like your "show me the code" approach :sweat_smile: Makes things much easier to understand.

mgi388:
Yeah I see the problem with the nitro example. If a player accidentally presses shift while driving and they do stop or slow down I could imagine that being frustrating for them.

Let's extend the driving example as a thought experiment. Imagine the developer has an action at the root context called `SaveGame` and they bind it to `Shift+S` (for argument's sake even though I know this may be a convoluted example). Well, if the player is driving and holding down `S`, then if they press `Shift`, I think that should trigger the save game action and it should them driving, right? Now, let's imagine that the developer does _not_ have the `SaveGame` action at all, and so there's no such `Shift+S` binding in the hierarchy. Would it be possible for BEI to be aware that in the context hierarchy there is some action with a modifier that should take priority over one without that modifier? With the `SaveGame` action in the root context being bound to `Shift+S`, could BEI work out that such an action exists when evaluating the `Move` action in the driving context?

Shatur:
This will be easy to implement, but logically it's a bit weird. `ModKeys` is a bitset, so we can't use 0 for `NONE` because that would represent an empty bitset. We could assign a bit to it, but that would make the internal logic a bit strange. `ModKeys` represents pressed modifiers. How would we handle `ModKeys::NONE | ModKeys::SHIFT`?

I'd probably suggest adding an option to `ActionSettings`. If it's set to `true`, check whether the exact number of modifiers are pressed (i.e., if there are no modifiers, no modifiers should be pressed). Basically, the logic you already implemented, just pass the boolean to the function. I don't think it's a *great* solution because users have to worry about it. But I don't think we can do much about it.

Still highly suggest to rethinking the approach. You can just move your action to another context.

mgi388:
Yeah no need to rush to implement it today, we can think on it and explore the design space first.

Supposing I tried to solve this with the existing context layering design, does it suggest _inverting_ some of the priorities? Or, at least, making a really high context for global things that shouldn't be overridden?

Revising the context layering example:

- **Player context** (priority 0):
  - `EnterCar` bound to `Enter`.
- **Driving context** (priority 1): 
  - `ExitCar` bound to `Enter` (overrides player context).
  - `ToggleMap` bound to `M`.
- **Global context** (priority 1000): 
  - `ToggleMute` bound to `Shift+M`. Because the priority is higher, `ToggleMap` never fires which is good—the player never observes the map to toggle when they were trying to toggle mute.

I tested this locally and I think this works as expected.

It feels a bit ugly to me that I have to make my global context some arbitrarily high number when I usually think of the first context I spawn as being the highest (imagine when the game loads and it has no concept of driving, you'd first spawn your global context and it should be the highest, which we could do if it had priority 0 which meant the highest). But I _think_ it works. Maybe there's other design issues doing this that I can't foresee.

WDYT?

PR #247, opened 2025-11-06
Add a helpful diagnostic when `IntoBindingBundle` is not implemented

janhohenheim:


Shatur (changes requested):
Awesome!

Noticed a small leftover from your crate :)

janhohenheim:
Whoops! Fixed

PR #262, opened 2025-12-25
Time kind default

Shatur:
I think it's a much better default.

Otherwise when you change your speed, things like camera controls will also work faster, which is not what user would expect. Or when you pause your game, things like `SmoothNudge` won't fade because your delta will be zero.

This only affect users who play with time dilation. But I target 0.18 branch since it's technically a breaking change and I plan to merge it only after 0.18 release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


alice-i-cecile (changes requested):
We should call out the default value explicitly in the TimeKind docs.

I'd also like a bit of discussion in those docs about the pros and cons of each choice, but that's bonus credit.

PR #263, opened 2025-12-26
Don't impose trait bounds on for `C` in `Actions<C>`

Shatur:


alice-i-cecile (approved):
We're missing `reflect` annotations for these traits too, but this is a straight improvement.

PR #269, opened 2026-01-09
Change default context evaluation order

Shatur:
Evaluate last-spawned contexts first. Modifiers and conditions use insertion order, but this behavior is less convenient for contexts.

For example, when spawning multiple dialogs with a `CloseDialog` action, closing should start from the most recent one.

Based on #262. ~Opening as a draft since it will target the next release after 0.18.~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Change default context evaluation order. Now last spawned context evaluated first.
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Change default context evaluation order. Now last spawned context evaluated first.
```
alice-i-cecile:
```suggestion
- Change default context evaluation order. Now the last spawned context is evaluated first.
```

alice-i-cecile (approved):

PR #279, opened 2026-01-29
Repro for: called `Result::unwrap()` on an `Err` value: QueryDoesNotMatch(…

mgi388:
…71v0, ArchetypeId(21))

- Run example: `cargo run --example context_layering`
- Press X
- Press Z
- Watch it crash

The example has been changed to disable some actions when you press X. Then when you press Z it despawns the context/player and you can see the crash.


Shatur:
Closing per #280. Just waiting for @alice-i-cecile review.

PR #280, opened 2026-01-29
Fix crash when despawning disabled entities

Shatur:
Conditions and modifiers didn't handle disabled entities. Closes #279.


mgi388 (approved):
LGTM. Can't replicate the bug anymore in my game or in the example.

alice-i-cecile (approved):

PR #282, opened 2026-01-30
Improve mocking ergonomics

janhohenheim:
TL;DR:
```rust
commands
  .entity(context)
  .mock_once::<Player, Jump>(ActionState::Fired, true);
```

### DIFF
--- a/tests/mocking.rs
+++ b/tests/mocking.rs
@@ -174,7 +174,6 @@ fn entity_command() {
 
     let context = app
         .world_mut()
-        // Using implicit `ActionMock`
         .spawn((TestContext, actions!(TestContext[Action::<Test>::new()])))
         .id();
 


tests/mocking.rs
```diff
@@ -40,6 +40,49 @@ fn updates() {
...
+    app.update();
+
+    let (&action, &state, _events) = actions.single(app.world()).unwrap();
+    assert!(!*action);
+    assert_eq!(state, ActionState::None);
+
+    app.world_mut()
+        .commands()
+        .entity(context)
+        .mock::<TestContext, Test>(ActionMock::once(ActionState::Fired, true));
```
mgi388: Just throwing it out there. I don't love the "mock" terminology and verb here. To me this reads like you're doing something with testing.

To keep "mock" terminology for now at least, maybe `mock_action` would be a better command name here:

```rust
app.world_mut()
        .commands()
        .entity(context)
        .mock_action::<TestContext, Test>(ActionMock::once(ActionState::Fired, true));
```

No pressure just a thought!

src/action/mock.rs
```diff
@@ -0,0 +1,272 @@
...
+/// (e.g., [`Start<A>`], [`Fire<A>`]) will still be triggered as usual.
+///
+/// Once the span expires, [`Self::enabled`] is set to `false`, and the action resumes
+/// the regular evaluation. The component is not removed automatically, allowing you
+/// to reuse it for future mocking.
+///
+/// Mocking does not take effect immediately - it is applied during the next context evaluation.
+/// For more details, see the [evaluation](crate#evaluation) section in the quick start guide.
+///
+/// See also [`ExternallyMocked`](crate::context::ExternallyMocked) to manually control the action data.
```
Shatur: I think it's better to cross-link things:
```suggestion
/// See also [`MockCommandExt`] for command-based API on the context entity and
/// [`ExternallyMocked`](crate::context::ExternallyMocked) to manually control the action data.
```

src/action/mock.rs
```diff
@@ -0,0 +1,272 @@
...
+    Manual,
+}
+
+impl From<Duration> for MockSpan {
+    fn from(value: Duration) -> Self {
+        Self::Duration(value)
+    }
+}
+
+/// Command used by [`MockCommandExt::mock`]. See that method for documentation.
```
Shatur: Let's simplify this:
```suggestion
/// Command used by [`MockCommandExt`].
```

src/action/mock.rs
```diff
@@ -0,0 +1,272 @@
...
+
+        *action_mock = self.action_mock;
+        Ok(())
+    }
+}
+
+/// Extension trait for [`EntityCommands`] that allows mocking actions.
+pub trait MockCommandExt {
+    /// Mocks an action. `C` is the action context, `A` is the [`InputAction`].
+    /// Convenience method so we don't have to manually query an action's [`ActionMock`].
```
Shatur: How about something like this?
```suggestion
    /// Searches for an entity with [`Action<A>`] in [`Actions<C>`] and inserts [`ActionMock`] to it with the given values.
    ///
    /// Convenience method to avoid manually querying an [`ActionMock`].
```

src/action/mock.rs
```diff
@@ -0,0 +1,272 @@
...
+    /// ```
+    fn mock<C: Component, A: InputAction + Send>(
+        &mut self,
+        state: ActionState,
+        value: impl Into<ActionValue>,
+        span: impl Into<MockSpan>,
+    ) -> &mut Self;
+
+    /// Mocks an action for a single update. `C` is the action context, `A` is the [`InputAction`].
+    /// Convenience method so we don't have to manually query an action's [`ActionMock`].
```
Shatur: How about to write something like this?

```suggestion
    /// Like [`Self::mock`], but uses [`MockSpan::Once`].
    ///
    /// Similar to [`ActionMock::once`].
```

This will make it obvious, so I think we will be able to remove the example. And if you agree, cold you also adjust the docstring for [`ActionMock::once`] the same way?

src/action/mock.rs
```diff
@@ -0,0 +1,260 @@
...
+#[derive(Reflect, Debug, Clone, Copy)]
+#[cfg_attr(feature = "serialize", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "serialize", reflect(Serialize, Deserialize))]
+pub enum MockSpan {
+    /// Active for a fixed number of context evaluations.
+    Updates(u32),
+    /// Active for a real-time [`Duration`].
+    Duration(Duration),
+    /// Remains active until [`ActionMock::enabled`] is manually set to `false`,
+    /// or the [`ActionMock`] component is removed from the action entity.
```
mgi388: Is this comment about "or the [`ActionMock`] component is removed from the action entity" invalid now that `ActionMock` is no longer checked to be `Option` in the system below?

As in, now that `ActionMock` is required, does that mean users shouldn't really ever be removing it?

Shatur: Adjusted the docs. I'm open to re-consider keeping `Option`, but I'd expect users to use the `enabled` field to avoid causing archetype moves.

Shatur (approved):

PR #283, opened 2026-01-30
Set ActionSettings::consume_input to false by default

janhohenheim:
Less footgunny that way :)

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -761,7 +761,7 @@ impl<C> Copy for ContextActivity<C> {}
 /// Ordering matters because actions may "consume" inputs, making them unavailable to other actions
 /// until the context that consumed them is evaluated again. This allows contexts layering, where
 /// some actions take priority over others. This behavior can be customized per-action by setting
-/// [`ActionSettings::consume_input`] to `false`.
+/// [`ActionSettings::consume_input`].
 ///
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Rename `ActionEvents::CANCELLED` into `ActionEvents::CANCEL`.
 - Rename `ActionEvents::COMPLETED` into `ActionEvents::COMPLETE`.
 - `ModKeys` now serializes to human-readable format instead of raw integers. Existing config files with integer `mod_keys` values need to be updated or regenerated.
+- `ActionSettings::consume_input` now defaults to `false`.
```
Shatur: Could you put it first? I think it's the most impactful breaking change. And how about to expand on it?
```suggestion
- `ActionSettings::consume_input` now defaults to `false`. This should make the input less footgunny. If you need to enable input consumption, you need to explicitly set this field to `true`.
```

Shatur (changes requested):
I think it's the right move.

Could you adjust line 762-763 in `context.rs` to something like this?
```
/// some actions take priority over others. This behavior can be customized per-action by setting
/// [`ActionSettings::consume_input`].
```

PR #290, opened 2026-02-04
Parity between mock API and existing command APIs

### LINKED ISSUE #289: Add `try_` versions of `EntityCommands` `mock` and `mock_once`
See [Discord](https://discord.com/channels/691052431525675048/1297361733886677036/1468565228538171412) for context.

Why? It's easy to experience the panic e.g., `entity 462v0 has no Actions<MyInputContext>` when you're spawning and despawning contexts (e.g., when using context layering) intertwined with mocking. 

Even if I set my Bevy error handler to warn, not panic, I'd still see warning logs telling me that a context doesn't exist, but there's nothing I can do about it, so having `try_` would be useful to callers can indicate they don't care if it fails.

janhohenheim:
API is now almost exactly like `remove` and `try_remove`, only thing I didn't bother porting was `remove_if` and `try_remove_if`.

This is teeeeechnically breaking, but honestly, I would publish this as a patch release.

Resolves #289

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- for parity with upstream methods, `mock` and `mock_once` commands now override the default error handler to emit a warning.
+- For parity with upstream methods, `mock` and `mock_once` commands now override the default error handler to emit a warning.
 
 ## [0.23.1] - 2026-02-03
 


mgi388 (approved):
Thanks for doing this.

src/action/mock.rs
```diff
@@ -203,6 +203,7 @@ pub trait MockEntityCommandsExt {
     /// Searches for an entity with [`Action<A>`] in [`Actions<C>`] and inserts [`ActionMock`] to it with the given values.
     ///
     /// Convenience method to avoid manually searching for an action on a context entity to insert [`ActionMock`].
+    /// This will emit a warning if the entity does not exist, does not have [`Actions<C>`], or those actions do not contain an [`Action<A>`].
```
Shatur: I think by default it will panic, right? Maybe we should adjust the wording to clarify that it depends on the configured error-handler?

Shatur (changes requested):
Thank you! Will publish a patch. I only have a tiny question about the wording 🙂

PR #293, opened 2026-02-05
feat: add Toggle condition

### LINKED ISSUE #291: Add `Toggle` condition
I've made a `Toggle` condition. Here's the gist: https://gist.github.com/mgi388/da68cb3e8fe42ee54326f9bb381a1966. 

See [Discord](https://discord.com/channels/691052431525675048/1297361733886677036/1457691842429587538) for more context.

Why? A toggle condition is useful for actions that you want to remain "on" after the player triggers them. For example, if the player presses `F` you can use a toggle to make the `SelectingFireTarget` action remain "on", and then when they press `F` again the action turns "off".

I haven't PRd this because a) I haven't added an updated example and b) I haven't ironed out kinks in using this in my own code. When I say kinks, I don't mean to say the implementation above is full of bugs because it's pretty a simple condition. It's just that I programmatically toggle the condition on and off and I haven't 100% ironed out all the bugs in my own code around all the other actions that need to play nicely with this.

Someone else is welcome to PR it, but I wanted to create the issue to share the gist publicly and also track the feature request.

niveshdandyan:
## Summary
Adds a `Toggle` condition that allows actions to remain "on" after being triggered, turning "off" when triggered again.

This is useful for actions that should persist after the player triggers them. For example, if the player presses `F` you can use a toggle to make the `SelectingFireTarget` action remain "on", and then when they press `F` again the action turns "off".

Fixes #291

## Changes
- Add `Toggle` condition implementation in `src/condition/toggle.rs`
- Register `Toggle` in the condition module exports
- Register `Toggle` with the input condition system in `EnhancedInputPlugin`
- Include unit tests for toggle behavior

## Implementation Details
The `Toggle` condition tracks:
- `actuation`: The trigger threshold (default: 0.5)
- `toggled`: Current toggle state (on/off)
- `actuated`: Internal state tracking previous actuation

The toggle state flips each time the input transitions from non-actuated to actuated. When toggled on, it returns `ActionState::Fired`; when toggled off, it returns `ActionState::None`.

Reference implementation: https://gist.github.com/mgi388/da68cb3e8fe42ee54326f9bb381a1966

## Testing
- [x] cargo check passes
- [x] cargo test passes (98 tests including new toggle test)
- [x] Toggle behavior works as expected:
  - Initially not actuated -> None
  - First actuation toggles on -> Fired
  - Continued actuation maintains -> Fired
  - Release (while toggled on) maintains -> Fired
  - Second actuation toggles off -> None
  - Third actuation toggles on again -> Fired

## AI Transparency
This PR was created with the assistance of AI (Claude by Anthropic) for code generation and review.

### DIFF
--- a/src/condition/toggle.rs
+++ b/src/condition/toggle.rs
@@ -1,0 +1,116 @@
+use bevy::prelude::*;
+
+use super::DEFAULT_ACTUATION;
+use crate::prelude::*;
+
+/// Returns [`ActionState::Fired`] when toggled on, [`ActionState::None`] when toggled off.
+///
+/// The toggle state flips each time the input transitions from non-actuated to actuated.
+/// This is useful for actions that should remain "on" after the player triggers them,
+/// and turn "off" when triggered again.
+///
+/// For example, if the player presses `F` you can use a toggle to make the `SelectingFireTarget`
+/// action remain "on", and then when they press `F` again the action turns "off".
+#[derive(Component, Reflect, Debug, Clone, Copy)]
+pub struct Toggle {
+    /// Trigger threshold.
+    pub actuation: f32,
+    /// Current toggle state (on/off).
+    pub toggled: bool,
+    actuated: bool,
+}
+
+impl Toggle {
+    #[must_use]
+    pub const fn new(actuation: f32) -> Self {
+        Self {
+            actuation,
+            toggled: false,
+            actuated: false,
+        }
+    }
+}
+
+impl Default for Toggle {
+    fn default() -> Self {
+        Self::new(DEFAULT_ACTUATION)
+    }
+}
+
+impl InputCondition for Toggle {
+    fn evaluate(
+        &mut self,
+        _actions: &ActionsQuery,
+        _time: &ContextTime,
+        value: ActionValue,
+    ) -> ActionState {
+        let previously_actuated = self.actuated;
+        self.actuated = value.is_actuated(self.actuation);
+
+        if self.actuated && !previously_actuated {
+            self.toggled = !self.toggled;
+        }
+
+        if self.toggled {
+            ActionState::Fired
+        } else {
+            ActionState::None
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::context;
+
+    #[test]
+    fn toggle() {
+        let (world, mut state) = context::init_world();
+        let (time, actions) = state.get(&world);
+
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -402,7 +402,7 @@ use context::{
     input_reader::{self, ConsumedInputs, PendingBindings},
 };
 use modifier::fns::ModifierRegistry;
-use prelude::{Press, Release, *};
+use prelude::{Press, Release, Toggle, *};
```
Shatur: Not needed:
```suggestion
use prelude::{Press, Release, *};
```

src/condition/toggle.rs
```diff
@@ -0,0 +1,116 @@
...
+use crate::prelude::*;
+
+/// Returns [`ActionState::Fired`] when toggled on, [`ActionState::None`] when toggled off.
+///
+/// The toggle state flips each time the input transitions from non-actuated to actuated.
+/// This is useful for actions that should remain "on" after the player triggers them,
+/// and turn "off" when triggered again.
+///
+/// For example, if the player presses `F` you can use a toggle to make the `SelectingFireTarget`
+/// action remain "on", and then when they press `F` again the action turns "off".
```
Shatur:
```suggestion
/// When the input is pressed:
/// - If currently off, turns on and fires continuously every frame
/// - If currently on, turns off and stops firing
///
/// This is useful for modes that should persist until toggled off,
/// like entering a "select target" mode, toggling crouch, or any other
/// action that represents a persistent state rather than a momentary input.
```

src/condition/toggle.rs
```diff
@@ -0,0 +1,116 @@
...
+    use crate::context;
+
+    #[test]
+    fn toggle() {
+        let (world, mut state) = context::init_world();
+        let (time, actions) = state.get(&world);
+
+        let mut condition = Toggle::default();
+
+        // Initially not actuated, should be None
```
Shatur: I prefer use panic messages instead of comments and avoid writing them for obvious things like this one. See [this](https://github.com/simgine/bevy_enhanced_input/issues/291#issuecomment-3850308270) message for details.

Shatur (changes requested):
I don't like AI-generated PRs because they can introduce subtle bugs that are statistically indistinguishable from regular code.

But since I want this feature, I inspected the PR carefully. The logic is identical to the linked gist (I diffed it). 

In additional to the requested changes, it needs a changelog entry.

PR #302, opened 2026-02-16
Add Toggle condition

### LINKED ISSUE #291: Add `Toggle` condition

jonas-meyer:
## Summary                                                                                                                                                                

Adds a `Toggle` condition that keeps an action firing after the input is released, toggling off on the next press. Useful for persistent modes like target selection or crouch.

Closes #291 

## Testing

  - `cargo test --lib condition::toggle` (2 tests)
  - `cargo test --doc condition::toggle` (1 doc test)
  - All CI checks pass locally (fmt, clippy, rustdoc, no-default-features)

### DIFF
--- a/src/condition/toggle.rs
+++ b/src/condition/toggle.rs
@@ -6,33 +6,12 @@ use crate::prelude::*;
 /// Returns [`TriggerState::Fired`] when toggled on, [`TriggerState::None`] when toggled off.
 ///
 /// When the input is pressed:
-/// - If currently off, turns on and fires continuously every frame
-/// - If currently on, turns off and stops firing
+/// - If currently off, turns on and fires continuously every frame.
+/// - If currently on, turns off and stops firing.
 ///
 /// This is useful for modes that should persist until toggled off,
 /// like entering a "select target" mode, toggling crouch, or any other
 /// action that represents a persistent state rather than a momentary input.
-///
-/// ```
-/// # use bevy::prelude::*;
-/// # use bevy_enhanced_input::prelude::*;
-/// # #[derive(Component)]
-/// # struct Player;
-/// # #[derive(InputAction)]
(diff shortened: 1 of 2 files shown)


src/condition/toggle.rs
```diff
@@ -0,0 +1,154 @@
...
+            TriggerState::None,
+        );
+        assert_eq!(
+            condition.evaluate(&actions, &time, 0.0.into()),
+            TriggerState::None,
+        );
+        assert_eq!(
+            condition.evaluate(&actions, &time, 1.0.into()),
+            TriggerState::Fired,
+        );
```
Shatur: I'd test it like this instead to make it symmetric.
```suggestion
        assert_eq!(
            condition.evaluate(&actions, &time, 1.0.into()),
            TriggerState::None,
        );
        assert_eq!(
            condition.evaluate(&actions, &time, 0.0.into()),
            TriggerState::None,
        );
```

PR #322, opened 2026-04-14
Added `Flick` Condition (for controllers)

jayhwkns:
I was checking out this crate for a game I wanted to make that has fighting-game style inputs and I'm pretty happy with what you made here! The only thing I felt was missing was a way to lock an action behind a flick (think like smash attacks in Smash Bros.). I took the liberty of adding it myself. It's a pretty simple implementation that only triggers when a stick crosses from a resting position to the actuation without lingering in the middle-zone for too long.

I tried to replicate the coding style of the other conditions and I added a single test that covers a few things, but if I missed anything or if my documentation wasn't specific enough, just let me know. If this feature isn't wanted, that's cool too, but I thought I'd offer my contribution.

```rust
            context.spawn((
                Action::<Kick>::new(),
                // Replicate feel of Smash Bros., which requires the
                // controller to flick in ~2 frames (1/30 second)
                // for certain actions
                Flick::new(0.0333).with_actuation(0.9),
                Bindings::spawn(Axial::left_stick()),
            ));
```


alice-i-cecile:
Oh that's a neat little helper. I'm down: this is reasonably common and would be more so if it was easier.

examples/marble.rs
```diff
@@ -0,0 +1,9 @@
+//! Demonstrates flicking functionality
```
Shatur: Looks like it was added by accident?

src/condition/flick.rs
```diff
@@ -0,0 +1,195 @@
...
+    /// The type of time used to advance the timer.
+    pub time_kind: TimeKind,
+
+    timer: Timer,
+
+    fired: bool,
+}
+
+impl Flick {
+    const DEFAULT_EXIT_ACTUATION: f32 = 0.3;
```
Shatur:
```suggestion
    pub const DEFAULT_EXIT_ACTUATION: f32 = 0.3;
```

src/condition/flick.rs
```diff
@@ -0,0 +1,195 @@
...
+/// to an actuation threshold in an amount of time set by the user, then
+/// [`TriggerState::Ongoing`] until the actuation threshold is exited.
+#[derive(Component, Debug, Clone)]
+#[cfg_attr(feature = "reflect", derive(Reflect), reflect(Clone, Component, Debug))]
+pub struct Flick {
+    /// The threshold the input must reach to activate
+    pub actuation: f32,
+
+    /// The threshold the input must exit to start the timer and the threshold
+    /// the stick must enter to begin a flick
```
Shatur:
```suggestion
    /// the stick must enter to begin a flick.
    ///
    /// By default it's set to [`Self::DEFAULT_EXIT_ACTUATION`].
```

src/condition/flick.rs
```diff
@@ -0,0 +1,195 @@
...
+use super::DEFAULT_ACTUATION;
+use crate::prelude::*;
+
+/// Returns [`TriggerState::Fired`] when the input transitions from a rest threshold
+/// to an actuation threshold in an amount of time set by the user, then
+/// [`TriggerState::Ongoing`] until the actuation threshold is exited.
+#[derive(Component, Debug, Clone)]
+#[cfg_attr(feature = "reflect", derive(Reflect), reflect(Clone, Component, Debug))]
+pub struct Flick {
+    /// The threshold the input must reach to activate
```
Shatur: In Rust all doc sentences should end with a dot.
```suggestion
    /// The threshold the input must reach to activate.
```

src/condition/flick.rs
```diff
@@ -0,0 +1,195 @@
+use bevy::prelude::*;
+
+use super::DEFAULT_ACTUATION;
+use crate::prelude::*;
+
+/// Returns [`TriggerState::Fired`] when the input transitions from a rest threshold
+/// to an actuation threshold in an amount of time set by the user, then
+/// [`TriggerState::Ongoing`] until the actuation threshold is exited.
```
Shatur: Not sure about the last part. I'd probably expect it to output `None` until you back to the rest position. And only after that you can flick again. What do you think about it?

src/condition/flick.rs
```diff
@@ -0,0 +1,195 @@
...
+    use crate::context;
+
+    #[test]
+    fn flick() {
+        let (mut world, mut state) = context::init_world();
+        let (time, actions) = state.get(&world);
+
+        let mut condition = Flick::new(0.5).with_actuation(0.9);
+
+        // Check successful flick
```
Shatur: I prefer to avoid leaving comments that explain something obvious. I'm trying to avoid duplicating logic in 2 languages (Rust and English), so I write comments only if I have something to add 🙂 
```suggestion
```

src/condition/flick.rs
```diff
@@ -0,0 +1,195 @@
...
+            .resource_mut::<Time<Real>>()
+            .advance_by(Duration::from_secs(1));
+        let (time, actions) = state.get(&world);
+
+        assert_eq!(
+            condition.evaluate(&actions, &time, 1.0.into()),
+            TriggerState::Ongoing,
+        );
+
+        // Check a flick that takes too long
```
Shatur: Maybe split this into a separate test? Call it `time_out`.

examples/kick_ball.rs
```diff
@@ -0,0 +1,159 @@
+//! Demonstrates flicking functionality for controllers.
```
Shatur: Usually I'm conservative about examples since they require more maintenance. Bevy changes rapidly + CI time (examples are quire slow to compile). So I usually don't add examples for modifiers.

But this one is fun! I want to keep it 😅
Maybe worth replacing `local_multiplayer` with this one, just spawn 2 balls? It's fine to ignore ball to ball collision just for the example. @alice-i-cecile what do you think?
@jayhwkns if you're not interested in adding a second player here and renaming it into `local_multiplayer` - it's fine, I'll merge it as is and do it later myself. I don't want to overwhelm you.

Shatur (changes requested):
Thank you a lot! Left a few suggestions 🙂

src/condition/flick.rs
```diff
@@ -0,0 +1,195 @@
...
+/// [`TriggerState::Ongoing`] until the actuation threshold is exited.
+#[derive(Component, Debug, Clone)]
+#[cfg_attr(feature = "reflect", derive(Reflect), reflect(Clone, Component, Debug))]
+pub struct Flick {
+    /// The threshold the input must reach to activate
+    pub actuation: f32,
+
+    /// The threshold the input must exit to start the timer and the threshold
+    /// the stick must enter to begin a flick
+    pub max_rest_threshold: f32,
```
Shatur: What do you think about this name?
```suggestion
    pub rest_threshold: f32,
```

src/condition/flick.rs
jayhwkns: The intended use case, again comparing to Smash Bros., is that you can use it to start sprinting. The Fire could represent the initial dash or dodge, and then you can stay sprinting until the user lets go of the stick. That's how I intended to use the condition for a game I was planning, anyways.

Besides, the `TriggerState::Ongoing` would just be unused without it. We might as well keep it just to give developers the option. Or else I can make this configurable by adding a public boolean field to `Flick`.

examples/marble.rs
jayhwkns: ack. Yes, it was. `marble.rs` was the original name of the `kick_ball.rs` example before I renamed it. I'll fix it when I get back and rebase my branch so it doesn't show up in the history.

src/condition/flick.rs
Shatur: Got it, let's keep it this way! I'd maybe adjust the description a little:

```suggestion
/// to an actuation threshold in specified amount of time, then
/// [`TriggerState::Ongoing`] until the actuation threshold is exited.
```

src/condition/flick.rs
jayhwkns: Understandable. I was just trying to block out the test case so it was easier to understand. I'll split into separate tests instead.

src/condition/flick.rs
jayhwkns: Yeah the name was previously clunky for sure. That works, though.

examples/kick_ball.rs
jayhwkns: Sure thing! I'll try to do that and resolve the other changes later today. Don't worry about overwhelming me, I'm happy to help.

jayhwkns:
@Shatur It should be all good now. Doc comments are improved, test cases are split up, and I combined `kick_ball` into `local_multiplayer`.

src/condition/flick.rs
Shatur: Hm... I'm looking at tests and I'm starting to doubt the current behavior.

When the input crosses `rest_threshold`, I'd expect it to be `Ongoing`, but we output `None`.
And after we output `Fire`, we output `Ongoing`, but I'd expect it to be `None`.
I.e. it's swapped to what I'd expect.

What if we swap this? If you want sprint after `Fire`, can you create a separate action with a lower priority (just put it after the flick one) and set flick to consume the input?

examples/local_multiplayer.rs
```diff
@@ -8,77 +8,90 @@ use bevy::{
     input::gamepad::{GamepadConnection, GamepadConnectionEvent},
     prelude::*,
 };
-use bevy_enhanced_input::prelude::*;
+use bevy_enhanced_input::prelude::{Press, *};
```
Shatur, line 11: What do you think about removing or reducing interpolation? Otherwise it takes quite some time to decelerate in order to kick again.

src/condition/flick.rs
```diff
@@ -0,0 +1,238 @@
...
+    /// The type of time used to advance the timer.
+    pub time_kind: TimeKind,
+
+    timer: Timer,
+
+    fired: bool,
+}
+
+impl Flick {
+    pub const DEFAULT_EXIT_ACTUATION: f32 = 0.3;
```
Shatur: Hm... Let's remove the constant entirely. Looks like it's used only in the constructor. Just mention that the value is 0.3 by default.

Shatur (changes requested):
Left 3 comments (one of them in the old thread about the behavior)

examples/local_multiplayer.rs
jayhwkns: Sorry, I don't understand. I don't think I am using interpolation at all here, unless maybe I am thinking of the wrong definition (lerping comes to mind for me). Kicking should be doable no matter what speed you're going and it appears to work that way whenever I try it.

I changed how input worked on keyboard a little bit to where you now need to "arm" a kick before the press of the movement key if that's what you're referring to?

src/condition/flick.rs
jayhwkns: Sure. I set it to 0.2 instead to match DeadZone's default since there's really no reason for them to be different imo.

examples/local_multiplayer.rs
Shatur: Ah, right (I should stop reviewing PRs at late nigh :sweat_smile:).
When you control the ball with the keyboard, it accelerates and decelerates a bit slow. That's what causes the behavior I'm referring to. What do you think about speeding it up a little?

Also after the ongoing change behavior, you no longer can move normally with gamepad. Maybe add an action for running like I suggested in https://github.com/simgine/bevy_enhanced_input/pull/322#discussion_r3089783070 ?

jayhwkns: Sure, I'll speed it up a little and add a speed cap to make turning around easier.

The change in ongoing behavior didn't affect anything when I tried it. After all, the example doesn't even use the ongoing trigger state. I'm still able to move without flicking just fine on both gamepad and keyboard. I'm curious why it's not working on your end.

alice-i-cecile (approved):
Well-made and a cool idea. I'm happy with this now :)

examples/local_multiplayer.rs
Shatur: > The change in ongoing behavior didn't affect anything when I tried it. 

Ah, you're right, I'm sorry. When I tried it with my gamepad, it looked like I can't control the ball, only flick. But looks like it's because of the low acceleration, so I didn't notice it. Yeah, I think if we tweak acceleration, it will be perfect.

So discard the last paragraph.

alice-i-cecile (changes requested):
Actually, one final thing: release notes please! You can find that in CHANGELOG.md for this repo.

Shatur (approved):
Made a few minor changes. Love the modifier and the new example. Thank you a lot!

jayhwkns:
Thank you, too! I haven't contributed much to open source before so this was a good learning experience for me.

Shatur:
Since adding a new modifier is not a breaking change, I'll draft a patch release tomorrow after work :slightly_smiling_face:

alice-i-cecile (approved):

Issue #307, opened 2026-02-27
Improve examples


Shatur:
We need to make the following changes to our examples to better teach users:
- [ ] Write a proper character controller example instead of the current one. See https://bevy.org/examples/movement/physics-in-fixed-timestep/ for more details. To put it short, it should run in `FixedUpdate` and accumulate input. Check how [`bevy_ahoy` does it](https://github.com/janhohenheim/bevy_ahoy/blob/90beac8e05a280acf2d61a654bf0fecbf3404b50/src/input.rs#L1).
- [ ] Add `pull_style_api` that mirrors `basic_action_management`, but uses pull-style API.
- [ ] Rework `local_multiplayer` showcases a character controller that runs in `Update`, which isn't nice (see the link above). Let's change it into a snake game with 2 snakes (to showcase local multiplayer). In this kind of game we can mention that it's ok to use `Update`. It also sounds fun 😅(offtopic, but I also recently discovered a [snake MMO via SSH](https://eieio.games/blog/secure-massively-multiplayer-snake/)).

Each change could be done in a separate PR. Contributions are welcome. I think it's a good opportunity for new contributors to play with the crate and Bevy in general 🙂

jayhwkns:
How do we feel about that third bullet point now that the `local_multiplayer` example was changed to something else? I just opened a PR setting up the fixed update for the character controller, so I can pretty easily go back and change that too (unless we really want to keep that one simple).

Shatur:
@jayhwkns I think it's worth keeping what you did and just adjust it to use fixed timestamp.

PR #323, opened 2026-04-19
Use a fixed timestep in the `character_controller` example

jayhwkns:
First bullet in #307 

Accumulates input and calculated physics in a fixed timestep like in [this example](https://bevy.org/examples/movement/physics-in-fixed-timestep/) and in [bevy_ahoy](https://github.com/janhohenheim/bevy_ahoy/blob/main/src/input.rs). It made no difference in feel for me, but from what I can tell this is supposed to make things smoother and prevent input from getting eaten in frame-drops. I tried to keep things simple.

I can also adjust the new `local_multiplayer` example to use a fixed timestep as well.

### DIFF
--- a/examples/character_controller.rs
+++ b/examples/character_controller.rs
@@ -19,8 +19,14 @@ fn main() {
         .add_plugins((DefaultPlugins, EnhancedInputPlugin))
         .add_input_context::<Player>()
         .add_systems(Startup, setup)
+        .init_resource::<DidFixedUpdateRunThisFrame>()
+        .add_systems(PreUpdate, reset_fixed_update_flag)
+        .add_systems(FixedPreUpdate, set_fixed_update_flag)
         .add_systems(FixedUpdate, calculate_physics)
-        .add_systems(RunFixedMainLoop, clear_input)
+        .add_systems(
+            RunFixedMainLoop,
+            clear_input.run_if(did_fixed_update_run_this_frame),
+        )
         .add_systems(FixedPostUpdate, apply_input)
         .add_observer(apply_movement)
         .add_observer(apply_jump)
@@ -141,4 +147,20 @@ struct Jump;
 struct AccumulatedInput {
     movement: f32,
     jump: bool,
+}
+
+// Fixed timestep boilerplate
+#[derive(Resource, Deref, DerefMut, Default)]
+struct DidFixedUpdateRunThisFrame(bool);
+
+fn reset_fixed_update_flag(mut flag: ResMut<DidFixedUpdateRunThisFrame>) {
+    flag.0 = false;
+}
+
+fn set_fixed_update_flag(mut flag: ResMut<DidFixedUpdateRunThisFrame>) {
+    flag.0 = true;
+}
+
+fn did_fixed_update_run_this_frame(flag: Res<DidFixedUpdateRunThisFrame>) -> bool {
+    flag.0
 }
--- a/examples/character_controller.rs
+++ b/examples/character_controller.rs
@@ -19,14 +19,11 @@ fn main() {
         .add_plugins((DefaultPlugins, EnhancedInputPlugin))
         .add_input_context::<Player>()
         .add_systems(Startup, setup)
-        .init_resource::<DidFixedUpdateRunThisFrame>()
-        .add_systems(PreUpdate, reset_fixed_update_flag)
-        .add_systems(FixedPreUpdate, set_fixed_update_flag)
+        .init_resource::<FixedUpdateRan>()
+        .add_systems(PreUpdate, reset_fixed_update_ran)
+        .add_systems(FixedPreUpdate, set_fixed_update_ran)
         .add_systems(FixedUpdate, calculate_physics)
-        .add_systems(
-            RunFixedMainLoop,
-            clear_input.run_if(did_fixed_update_run_this_frame),
-        )
+        .add_systems(RunFixedMainLoop, clear_input.run_if(fixed_update_ran))
         .add_systems(FixedPostUpdate, apply_input)
         .add_observer(apply_movement)
         .add_observer(apply_jump)
@@ -151,16 +148,16 @@ struct AccumulatedInput {
 
 // Fixed timestep boilerplate
 #[derive(Resource, Deref, DerefMut, Default)]
-struct DidFixedUpdateRunThisFrame(bool);
+struct FixedUpdateRan(bool);
 
-fn reset_fixed_update_flag(mut flag: ResMut<DidFixedUpdateRunThisFrame>) {
-    flag.0 = false;
+fn reset_fixed_update_ran(mut ran: ResMut<FixedUpdateRan>) {
+    **ran = false;
 }
 
-fn set_fixed_update_flag(mut flag: ResMut<DidFixedUpdateRunThisFrame>) {
-    flag.0 = true;
+fn set_fixed_update_ran(mut ran: ResMut<FixedUpdateRan>) {
+    **ran = true;
 }
 
-fn did_fixed_update_run_this_frame(flag: Res<DidFixedUpdateRunThisFrame>) -> bool {
-    flag.0
+fn fixed_update_ran(ran: Res<FixedUpdateRan>) -> bool {
+    **ran
 }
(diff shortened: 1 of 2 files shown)


examples/character_controller.rs
```diff
@@ -63,25 +70,45 @@ fn setup(
...
-fn apply_movement(movement: On<Fire<Movement>>, mut query: Query<&mut PlayerPhysics>) {
-    let mut physics = query.get_mut(movement.context).unwrap();
-    physics.velocity.x = movement.value;
+fn apply_movement(
+    movement: On<Fire<Movement>>,
+    mut accumulated_inputs: Query<&mut AccumulatedInput>,
+) {
+    if let Ok(mut accumulated_inputs) = accumulated_inputs.get_mut(movement.context) {
+        accumulated_inputs.movement = movement.value;
+    }
```
Shatur: May I ask you to keep unwraps?
I use them for "this never happens and it's obvious why" and I feel like this example fits this case.

examples/character_controller.rs
```diff
@@ -63,25 +70,45 @@ fn setup(
...
 
-fn apply_jump(jump: On<Fire<Jump>>, mut query: Query<&mut PlayerPhysics>) {
-    let mut physics = query.get_mut(jump.context).unwrap();
-    if physics.is_grounded {
-        // Jump only if on the ground.
-        physics.velocity.y = JUMP_VELOCITY;
-        physics.is_grounded = false;
+fn apply_jump(jump: On<Fire<Jump>>, mut accumulated_inputs: Query<&mut AccumulatedInput>) {
+    if let Ok(mut accumulated_inputs) = accumulated_inputs.get_mut(jump.context) {
+        accumulated_inputs.jump = true;
```
Shatur: `accumulated_inputs` is a bit too long as for my personal taste:
```suggestion
fn apply_jump(jump: On<Fire<Jump>>, mut inputs: Query<&mut AccumulatedInput>) {
    if let Ok(mut input) = inputs.get_mut(jump.context) {
        accumulated_inputs.jump = true;
```

examples/character_controller.rs
```diff
@@ -63,25 +70,45 @@ fn setup(
...
 }
 
-fn calculate_physics(time: Res<Time>, mut query: Query<(&mut Transform, &mut PlayerPhysics)>) {
+fn clear_input(mut accumulated_inputs: Query<&mut AccumulatedInput>) {
+    for mut accumulated_input in &mut accumulated_inputs {
+        *accumulated_input = Default::default();
+    }
+}
+
+fn run_character_controller(players: Query<(&mut PlayerPhysics, &AccumulatedInput)>) {
```
Shatur:
```suggestion
fn apply_input(players: Query<(&mut PlayerPhysics, &AccumulatedInput)>) {
```

examples/character_controller.rs
```diff
@@ -63,25 +70,45 @@ fn setup(
...
+        if input.jump && physics.is_grounded {
+            physics.velocity.y = JUMP_VELOCITY;
+            physics.is_grounded = false;
+        }
+    }
+}
+
+fn calculate_physics(
+    fixed_time: Res<Time<Fixed>>,
+    mut query: Query<(&mut Transform, &mut PlayerPhysics)>,
```
Shatur: For consistency (I like `players` that you used instead of `query`):
```suggestion
    mut players: Query<(&mut Transform, &mut PlayerPhysics)>,
```

Shatur (changes requested):
Thank you! Looks good, left a few style suggestions.

> I can also adjust the new local_multiplayer example to use a fixed timestep as well.

It would be nice!

Shatur:
Also CC @janhohenheim.

jayhwkns:
On that last commit:

Turns out I technically missed some extra boilerplate needed to get the fixed timestep working properly. It wasn't noticeable in the character controller example, but the kick action would sometimes be prematurely cleared without the boilerplate. Seeing as it took just a copy and paste to get it working and it is required for this pattern, maybe it would be good to make this boilerplate a library feature? Maybe even a part of the DefaultPlugins Bevy provides.

examples/character_controller.rs
```diff
@@ -113,3 +141,26 @@ struct Movement;
...
+/// Accumulated input since the last fixed update.
+#[derive(Component, Default)]
+struct AccumulatedInput {
+    movement: f32,
+    jump: bool,
+}
+
+// Fixed timestep boilerplate
+#[derive(Resource, Deref, DerefMut, Default)]
+struct DidFixedUpdateRunThisFrame(bool);
```
Shatur:
```suggestion
struct FixedUpdateRan(bool);
```

examples/character_controller.rs
```diff
@@ -113,3 +141,26 @@ struct Movement;
...
+    flag.0 = false;
+}
+
+fn set_fixed_update_flag(mut flag: ResMut<DidFixedUpdateRunThisFrame>) {
+    flag.0 = true;
+}
+
+fn did_fixed_update_run_this_frame(flag: Res<DidFixedUpdateRunThisFrame>) -> bool {
+    flag.0
+}
```
Shatur: I'd suggest this naming. I know it's from the Bevy example, but I like this more 🙂 
```suggestion
fn reset_fixed_update_ran(mut ran: ResMut<FixedUpdateRan>) {
    **ran = false;
}

fn set_fixed_update_ran(mut ran: ResMut<FixedUpdateRan>) {
    **ran = true;
}

fn fixed_update_ran(ran: Res<FixedUpdateRan>) -> bool {
    **ran
}
```

Shatur (changes requested):
> maybe it would be good to make this boilerplate a library feature?

Definitely! Worth opening an issue.

examples/character_controller.rs
jayhwkns: I like that better, too. At least for the purpose of the example. I think the more verbose name might be better if this does end up being a library feature, just for clarity of it being *this frame* and all.

Shatur: Even when upstreamed, I'd put the more detailed information in the doc comment instead of the name.

examples/character_controller.rs
```diff
@@ -63,25 +73,40 @@ fn setup(
...
+            physics.is_grounded = false;
+        }
+    }
+}
+
+fn calculate_physics(
+    fixed_time: Res<Time<Fixed>>,
+    mut players: Query<(&mut Transform, &mut PlayerPhysics)>,
+) {
+    for (mut transform, mut physics) in players.iter_mut() {
```
Shatur:
```suggestion
    for (mut transform, mut physics) in &mut players {
```

examples/character_controller.rs
```diff
@@ -1,5 +1,9 @@
 //! Demonstrates how to create a simple platforming 2D character controller
 //! using actions with both keyboard and gamepad controls.
+//!
+//! Input is accumulated and applied to physics in a fixed timestep, as
+//! recommended in [this Bevy example](https://bevy.org/examples/movement/physics-in-fixed-timestep/)
+//! and as used in [bevy_ahoy](https://github.com/janhohenheim/bevy_ahoy).
```
Shatur, line 6: What do you think about moving the comment to the `AccumulatedInput`? And mention that it's needed specifically for character controllers.

Shatur (changes requested):
Great! One final review pass

examples/character_controller.rs
jayhwkns: It was my interpretation that we wanted to make this fixed timestep change to encourage the users of this crate to follow this practice, so I put the comment at the top of the module to emphasize its importance.

What about something like this?

```suggestion
//! For kinematic character controllers, input should be accumulated and applied to physics in a fixed timestep as
//! recommended in [this Bevy example](https://bevy.org/examples/movement/physics-in-fixed-timestep/)
//! and as used in [bevy_ahoy](https://github.com/janhohenheim/bevy_ahoy).
```

Shatur: Okay, let's keep it top module!

I like it!

Shatur (approved):

alice-i-cecile (approved):

PR #327, opened 2026-04-30
Simplified AccumulatedInput clearing in examples

jayhwkns:
Turns out I was wrong on [this comment](https://github.com/simgine/bevy_enhanced_input/pull/323#issuecomment-4276564250) on my last PR.

It's true that both [this example](https://bevy.org/examples/movement/physics-in-fixed-timestep/) and bevy_ahoy use a similar `did_fixed_timestep_run_this_frame` mechanism, but both these cases are meant for greater extensibility and looser ordering. For the case of these examples, its fine just to chain the systems together.

As long as character controllers follow the pattern of:

accumulate input on input events -> apply input -> clear input

in that order, no input should be lost.

### DIFF
--- a/examples/local_multiplayer.rs
+++ b/examples/local_multiplayer.rs
@@ -27,9 +27,8 @@ fn main() {
         .add_systems(
             FixedUpdate,
             (
-                (apply_input, clear_input).chain(),
+                (apply_input, calculate_physics, clear_input).chain(),
                 update_gamepads,
-                calculate_physics,
             ),
         )
         .add_observer(apply_roll)
--- a/examples/local_multiplayer.rs
+++ b/examples/local_multiplayer.rs
@@ -23,13 +23,15 @@ fn main() {
     App::new()
         .add_plugins((DefaultPlugins, EnhancedInputPlugin))
         .add_input_context::<Player>()
-        .init_resource::<FixedUpdateRan>()
-        .add_systems(PreUpdate, reset_fixed_update_ran)
-        .add_systems(FixedPreUpdate, set_fixed_update_ran)
         .add_systems(Startup, setup)
-        .add_systems(FixedUpdate, (calculate_physics, update_gamepads))
-        .add_systems(RunFixedMainLoop, clear_input.run_if(fixed_update_ran))
-        .add_systems(FixedPostUpdate, apply_input)
+        .add_systems(
+            FixedUpdate,
+            (
+                (apply_input, clear_input).chain(),
+                update_gamepads,
+                calculate_physics,
+            ),
+        )
         .add_observer(apply_roll)
         .add_observer(apply_kick)
         .run();
@@ -276,20 +278,4 @@ struct ArmKick;
 struct AccumulatedInput {
     roll: Vec2,
     kick: Option<Vec2>,
-}
-
-// Fixed timestep boilerplate
-#[derive(Resource, Deref, DerefMut, Default)]
-struct FixedUpdateRan(bool);
-
-fn reset_fixed_update_ran(mut ran: ResMut<FixedUpdateRan>) {
-    **ran = false;
-}
-
-fn set_fixed_update_ran(mut ran: ResMut<FixedUpdateRan>) {
-    **ran = true;
-}
-
-fn fixed_update_ran(ran: Res<FixedUpdateRan>) -> bool {
-    **ran
 }
--- a/examples/character_controller.rs
+++ b/examples/character_controller.rs
@@ -19,11 +19,10 @@ fn main() {
         .add_plugins((DefaultPlugins, EnhancedInputPlugin))
         .add_input_context::<Player>()
         .add_systems(Startup, setup)
-        .init_resource::<FixedUpdateRan>()
-        .add_systems(PreUpdate, reset_fixed_update_ran)
-        .add_systems(FixedPreUpdate, set_fixed_update_ran)
-        .add_systems(FixedUpdate, calculate_physics)
-        .add_systems(RunFixedMainLoop, clear_input.run_if(fixed_update_ran))
+        .add_systems(
+            FixedUpdate,
+            ((calculate_physics), (apply_input, clear_input).chain()),
+        )
         .add_systems(FixedPostUpdate, apply_input)
         .add_observer(apply_movement)
         .add_observer(apply_jump)
@@ -144,20 +143,4 @@ struct Jump;
 struct AccumulatedInput {
     movement: f32,
     jump: bool,
-}
-
-// Fixed timestep boilerplate
-#[derive(Resource, Deref, DerefMut, Default)]
-struct FixedUpdateRan(bool);
-
-fn reset_fixed_update_ran(mut ran: ResMut<FixedUpdateRan>) {
-    **ran = false;
-}
-
-fn set_fixed_update_ran(mut ran: ResMut<FixedUpdateRan>) {
-    **ran = true;
-}
-
-fn fixed_update_ran(ran: Res<FixedUpdateRan>) -> bool {
-    **ran
 }
--- a/examples/character_controller.rs
+++ b/examples/character_controller.rs
@@ -21,7 +21,7 @@ fn main() {
         .add_systems(Startup, setup)
         .add_systems(
             FixedUpdate,
-            ((calculate_physics), (apply_input, clear_input).chain()),
+            (apply_input, calculate_physics, clear_input).chain(),
         )
         .add_observer(apply_movement)
         .add_observer(apply_jump)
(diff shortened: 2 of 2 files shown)


examples/local_multiplayer.rs
```diff
@@ -23,13 +23,15 @@ fn main() {
...
-        .add_systems(PreUpdate, reset_fixed_update_ran)
-        .add_systems(FixedPreUpdate, set_fixed_update_ran)
         .add_systems(Startup, setup)
-        .add_systems(FixedUpdate, (calculate_physics, update_gamepads))
-        .add_systems(RunFixedMainLoop, clear_input.run_if(fixed_update_ran))
-        .add_systems(FixedPostUpdate, apply_input)
+        .add_systems(
+            FixedUpdate,
+            (
+                (apply_input, clear_input).chain(),
```
Shatur: Quick question: why we don't chain `calculate_physics` and `apply_input`?

jayhwkns: `calculate_physics` applies velocity to transform and handles friction and collisions, so it is mostly independent of input. I guess that it would probably make sense run `apply_input` first for responsiveness, though. Good catch. I just fixed it.

alice-i-cecile (approved):

examples/local_multiplayer.rs
```diff
@@ -23,13 +23,14 @@ fn main() {
...
-        .add_systems(FixedPreUpdate, set_fixed_update_ran)
         .add_systems(Startup, setup)
-        .add_systems(FixedUpdate, (calculate_physics, update_gamepads))
-        .add_systems(RunFixedMainLoop, clear_input.run_if(fixed_update_ran))
-        .add_systems(FixedPostUpdate, apply_input)
+        .add_systems(
+            FixedUpdate,
+            (
+                (apply_input, calculate_physics, clear_input).chain(),
+                update_gamepads,
```
Shatur, line 31: Why `update_gamepads` is a part of the `FixedUpdate`? I know it was like this before, but I think it should run in `PreUpdate` right after the Bevy input system. What do you think?

Shatur (changes requested):
One last suggestion

Shatur:
@jayhwkns I talked with @janhohenheim and he raised a few concerns.
For example, if multiple fixed updates happen in one frame, but you clear the inputs in the fixed updates themselves, the second fixed update that frame won't have any input. So e.g. if you have a game where you need to hold space to jump higher, you will randomly jump short since the input was interrupted. Or you will slow down for one tick, since only the first tick that frame has the W key held.

He is planning to take a look at the examples and fix the remaining oversights / add some comments soon.

examples/local_multiplayer.rs
jayhwkns: I'm pretty indifferent and was just keeping it how it was. In my opinion it really doesn't matter because its never gonna be dire to have a controller connect or disconnect at a certain time in the schedule.

Shatur: Updating it during the fixed update is not useful, since if the fixed update runs more than once, you'll run a system for nothing: it won't have any new events. It's usually better to update it as soon as possible, i.e. right after the gamepad event.

mgi388:
@janhohenheim if you haven't already, you could codify this into a test(s) of your `DidFixedTimestepRunThisFrame` design (not necessarily in BEI, just somewhere). That means you don't have to solely rely on comments to explain this and you could just refer people to the test(s) to explain why it has to be this way.

jayhwkns:
> @jayhwkns I talked with @janhohenheim and he raised a few concerns. For example, if multiple fixed updates happen in one frame, but you clear the inputs in the fixed updates themselves, the second fixed update that frame won't have any input. So e.g. if you have a game where you need to hold space to jump higher, you will randomly jump short since the input was interrupted. Or you will slow down for one tick, since only the first tick that frame has the W key held.
> 
> He is planning to take a look at the examples and fix the remaining oversights / add some comments soon.

Ah, I see the difference now. I was staring at how the example from #307 schedule was different from chaining and it wasn't clicking. Correct me if I'm wrong, but the difference is kind of like this?

*Correct way that bevy_ahoy and the example do it*
```
fixed-main-loop():
  for each fixed-timestep-run-this-frame:
     // do stuff
  clear-input() (after)
```

*How I did it in this PR*
```
fixed-main-loop():
  for each fixed-timestep-run-this-frame:
    // do stuff AND
    clear-input() (each time)
```

I suppose my concern would still be that running multiple fixed timesteps in the FixedMainLoop wouldn't accumulate different input values, but would just apply the same accumulated input each time? I don't know for sure as I haven't looked that deeply into the engine and how its scheduler works.

If we like the old way but want to cut down on boilerplate, I made a [crate](https://github.com/jayhwkns/bevy_schedule_utils) today that aims to reduce the boilerplate a little.

janhohenheim:
@jayhwkns exactly, you got it :)
Yes, it just applies the full input each time. A sophisticated algorithm would use input timestamps to figure out *how much* of a given input it should apply in each frame, and scale the input down appropriately. We can approximate this today using `Time<Real>` and `Time<Fixed>`, but a fully robust version would use windowing timestamps. Which `winit` does not provide, bleh.

jayhwkns:
@janhohenheim I see. I was wondering why bevy_ahoy (awesome work, btw) had all those Stopwatches in AccumulatedInput. I assume this is why?

I suppose at this point the question is how complex vs how accurate we want these two examples to be? It sounds like we could get deeper into a rabbit hole and make a simple example much more intimidating than need be. Maybe it would be good to concede the limits of the example in a doc comment and link a resource to a more robust guide on character controllers?

Either way, I believe this PR's changes might not result in any different behavior than there was before. It all depends on where exactly `clear_input` is placed in `RunFixedMainLoop` and [the docs](https://docs.rs/bevy/latest/bevy/prelude/struct.Main.html) don't exactly make it clear when in the loop it would be run (or if its nondeterministically just somewhere during the fixed main loop).

janhohenheim:
Not quite. The stopwatches are there to know how long it has been since the input was last issued. That's for allowing some leeway when the player pressed something slightly too early or too late. For example, they are allowed to press the jump button a few milliseconds before they actually touch the ground and the KCC will jump as soon as they touch it. Same principle as with coyote time (allow jumping even a bit after after leaving a platform), just the other way around.

I would want an example for character controllers to be accurate, since IME many people are copy-pasting character-controller related code and then asking on Discord why their controller behaves wrong, at which point they learn that they need to refactor the entire thing. That said, I don't think we need to add stopwatches here, nor timestamp based accumulation. Just replaying the inputs for each fixed update is *fine* in most cases, if a bit inaccurate, but it won't lead to any bugs.

To schedule the clear correctly, you place it in `RunFixedMainLoop` and `.in_set(FixedMainLoopSystems::AfterFixedMainLoop)`, which will run every single frame after the 0-n fixed updates happened. Combined with `.run_if(fixed_update_did_happen)`, that clear will only happen after 1-n fixed updates. Yes, I know, the system set naming is horrible here and the docs are a bit confusing. I wonder who gave them those terrible names (it was me :D)

jayhwkns:
Closing this PR to shift the goal post based on feedback. This fix isn't correct, but neither is the original scheduling I implemented. Gonna focus more on correctness instead of simplicity. I'm trying to write a character controller crate alongside this, so I want to ensure that everything is right. Thank you all for the help!

jayhwkns:
> @jayhwkns exactly, you got it :) Yes, it just applies the full input each time. A sophisticated algorithm would use input timestamps to figure out _how much_ of a given input it should apply in each frame, and scale the input down appropriately. We can approximate this today using `Time<Real>` and `Time<Fixed>`, but a fully robust version would use windowing timestamps. Which `winit` does not provide, bleh.

@janhohenheim I tried to spend some time racking my brain around what you said here. I couldn't find anything using `Time<Real>` to approximate this, so I tried modeling it out myself.

In this scenario, the framerate is much lower than the 64Hz fixed update. In bevy_enhanced_input, as far as I can tell, the input is only read on `PreUpdate` in the `EnhancedInputSystems::Prepare` set, meaning it remains constant until the next frame update. In the current implementation, applied input is scaled by fixed delta time, so the original issue of applying the same input over and over isn't as big of a problem as I thought considering that it is inevitable that the input can't change between fixed update with the way that it is currently polled.

image

As the red marks point out in the diagram, there is a scenario where the new input is applied for a full fixed-delta-t despite being input for a brief moment. Depending on your perspective, this could be a good thing since the user might get some extra compensation for the delay resulting from reading the input, but it might not be fully accurate.

A potential solution to this would be to retain the previously accumulated input upon clearing, then detect this scenario when `fixed dt` > `real dt`, then applying the old input scaled by `fixed dt - real dt` and the new input scaled by `real dt`.

Is this what you meant at all? Maybe I am overthinking it, but let me know what you think.

PR #332, opened 2026-05-04
Changed `DeltaScale` default to virtual time, added explicit constructors

### LINKED ISSUE #331: Change `DeltaScale` to use `TimeKind::Virtual` as default
I think it may be more practical to use Virtual time as the default for `DeltaScale`. 

Picture a typical use case: you are using `DeltaScale::default()` with the left stick to accelerate the player. The change in velocity is scaled by real time, but we'll say that the game has a modifier that's slowing the game down to 0.5x speed. The player's movement speed will be affected, but they will accelerate much 2x faster in in-game time.

Or in possibly a more common case, the user doesn't have a check to see if the game is paused in the system that handles movement input. They hold left on the stick while in the pause screen and unpause the game. They accelerate immediately without any checks.

I haven't ran into these issues personally, but I can definitely picture someone spending a while trying to find the issue if this were to happen to them.

I agree that the default `TimeKind` being `Real` makes sense in a lot of cases. You might want the player to still just hold a button for the same amount of real time to fire and action regardless of if the game is slowed down, but for DeltaScale I believe it should be handled differently.

Alternatively, we could move this choice onto the user explicitly by replacing the `Default` implementation with functions like `DeltaScale::real()` and `DeltaScale::virtual()`.

jayhwkns:
CLOSE #331 

Name is self explanatory. I've confirmed that the pause bug in the issue was in fact present when a pause screen was introduced (and `=` was changed to `+=` to properly accumulate the input). See for yourself by setting the `DeltaScale` in `local_multiplayer.rs` to `real_time()`, then hold a direction while paused and see the ball slingshot when unpausing.

A few minor improvements for testing, though it is still quite difficult to simulate the effects of pausing and scaling virtual time in unit tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - `Chord::ongoing` to control whether partial activation returns `TriggerState::Ongoing` or `TriggerState::None`.
+- `DeltaScale::real_time()` and `DeltaScale::virtual_time()` constructors.`
 
 ### Removed
 
@@ -18,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Improve action evaluation logging.
+- **Breaking**: `DeltaScale::default()` now uses virtual time instead of real time.
 
 ## [0.24.4] - 2026-04-18
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Improve action evaluation logging.
-- **Breaking**: `DeltaScale::default()` now uses virtual time instead of real time.
+- `DeltaScale::default()` now uses virtual time instead of real time. This is most likely what you want, since it's equivalent to getting the delta from [the default `Time` resource](https://docs.rs/bevy/latest/bevy/prelude/struct.Time.html). If you want the original behavior, use `DeltaScale::real_time()`.
 
 ## [0.24.4] - 2026-04-18
 
--- a/examples/character_controller.rs
+++ b/examples/character_controller.rs
@@ -65,9 +65,8 @@ fn setup(
             (
                 Action::<Movement>::new(),
                 DeadZone::default(),
+                // A smoothed input is directly applied to velocity.
                 SmoothNudge::default(),
-                // We don't apply `DeltaScale` here because the movement vector is
-                // multiplied by delta time during the physics calculation.
                 Scale::splat(450.0),
                 Bindings::spawn((
                     Bidirectional::new(KeyCode::KeyD, KeyCode::KeyA),
--- a/examples/character_controller.rs
+++ b/examples/character_controller.rs
@@ -65,8 +65,10 @@ fn setup(
             (
                 Action::<Movement>::new(),
                 DeadZone::default(),
-                // A smoothed input is directly applied to velocity.
+                // Velocity is set to a smoothed input to simulate acceleration.
                 SmoothNudge::default(),
+                // We don't apply `DeltaScale` here because the movement vector is
+                // multiplied by delta time during the physics calculation.
                 Scale::splat(450.0),
                 Bindings::spawn((
                     Bidirectional::new(KeyCode::KeyD, KeyCode::KeyA),
--- a/examples/local_multiplayer.rs
+++ b/examples/local_multiplayer.rs
@@ -25,6 +25,7 @@ fn main() {
     App::new()
         .add_plugins((DefaultPlugins, EnhancedInputPlugin))
         .add_input_context::<Player>()
+        .add_input_context::<Shared>()
         .init_resource::<FixedUpdateRan>()
         .add_systems(Startup, setup)
         .add_systems(PreUpdate, (reset_fixed_update_ran, update_gamepads))
@@ -39,6 +40,7 @@ fn main() {
         )
         .add_observer(accumulate_roll)
         .add_observer(accumulate_kick)
+        .add_observer(toggle_pause)
         .run();
 }
 
@@ -98,6 +100,7 @@ fn setup(
         Transform::from_xyz(-80.0, 0.0, 0.0),
     ));
 
+    // Player 2
     let material2 = materials.add(Color::srgb(0.9, 0.1, 0.1));
     commands.spawn(player_bundle(
         Player::Second,
@@ -105,6 +108,32 @@ fn setup(
         ball_mesh,
         material2,
         Transform::from_xyz(80.0, 0.0, 0.0),
+    ));
+
+    // Shared player controls
+    commands.spawn((
+        Shared,
+        actions!(
+            Shared[(
+                Action::<Pause>::new(),
+                Press::new(1.0),
+                bindings![KeyCode::Escape, GamepadButton::Start]
+            )]
+        ),
+    ));
+
+    // Pause indicator
+    commands.spawn((
+        PauseIndicator,
+        Node {
+            width: percent(100),
+            height: percent(100),
+            align_items: AlignItems::Center,
+            justify_content: JustifyContent::Center,
+            display: Display::None,
+            ..default()
+        },
+        children![Text::new("Paused")],
     ));
 }
 
@@ -167,7 +196,7 @@ fn player_bundle(
             context.spawn((
                 Action::<Roll>::new(),
                 DeadZone::default(),
-                DeltaScale::default(),
+                DeltaScale::virtual_time(),
                 Scale::splat(ACCELERATION),
                 Bindings::spawn(dir_bindings),
             ));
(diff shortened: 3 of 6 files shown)


alice-i-cecile (changes requested):
This needs a release note, which needs to make it clear that this is a breaking change :)

examples/character_controller.rs
```diff
@@ -65,9 +65,8 @@ fn setup(
             (
                 Action::<Movement>::new(),
                 DeadZone::default(),
+                // A smoothed input is directly applied to velocity.
                 SmoothNudge::default(),
-                // We don't apply `DeltaScale` here because the movement vector is
-                // multiplied by delta time during the physics calculation.
```
Shatur: Why remove the comment?

src/context/time.rs
```diff
@@ -10,7 +10,7 @@ use bevy::{ecs::system::SystemParam, prelude::*};
 #[derive(SystemParam, Deref)]
 pub struct ContextTime<'w> {
     #[deref]
-    pub virt: Res<'w, Time>,
+    pub virt: Res<'w, Time<Virtual>>,
```
Shatur: Why switch this `Virtual`? This will break the time in the fixed schedule. See [the documentation](https://docs.rs/bevy/latest/bevy/prelude/struct.Time.html).

If you think `TimeKind::Virtual` is confusing, but we can rename it into `TimeKind::Current` or `TimeKind::Default`.

src/modifier/delta_scale.rs
```diff
@@ -5,17 +5,38 @@ use crate::prelude::*;
...
+
+    pub fn virtual_time() -> Self {
+        Self {
+            time_kind: TimeKind::Virtual,
+        }
+    }
+}
+
+impl Default for DeltaScale {
+    /// Uses `TimeKind::Virtual` by default.
```
Shatur: I'd put the comment on the field.

src/modifier/delta_scale.rs
```diff
@@ -51,31 +72,61 @@ mod tests {
     use crate::context;
 
     #[test]
-    fn scaling() {
+    fn real_scaling() {
```
Shatur: I don't think we need tests for this. It picks the correct resource based on what we choose. It would be tedious to test this in every modifier and condition that reads time. If we want to test the timer behavior, it should be on the `ContextTime`, but I don't think it needs tests at all.

CHANGELOG.md
```diff
@@ -18,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Improve action evaluation logging.
+- **Breaking**: `DeltaScale::default()` now uses virtual time instead of real time.
```
Shatur:
```suggestion
- `DeltaScale::default()` now uses virtual time instead of real time. This is most likely what you want, since it's equivalent to getting the delta from [the default `Time` resource](https://docs.rs/bevy/latest/bevy/prelude/struct.Time.html). If you want the original behavior, use `DeltaScale::real_time()`.
```

src/modifier/delta_scale.rs
```diff
@@ -5,17 +5,38 @@ use crate::prelude::*;
...
+        Self {
+            time_kind: TimeKind::Real,
+        }
+    }
+
+    pub fn virtual_time() -> Self {
+        Self {
+            time_kind: TimeKind::Virtual,
+        }
+    }
```
Shatur: Maybe turn into constants?

examples/local_multiplayer.rs
```diff
@@ -256,6 +285,21 @@ fn advance_physics(time: Res<Time>, players: Query<(&mut Transform, &mut PlayerP
     }
 }
 
+fn toggle_pause(
+    _: On<Fire<Pause>>,
```
Shatur: For consistency:
```suggestion
    _on: On<Fire<Pause>>,
```

examples/local_multiplayer.rs
```diff
@@ -285,6 +329,16 @@ struct AccumulatedInput {
...
 
+#[derive(Component)]
+struct Shared;
+
+#[derive(InputAction)]
+#[action_output(bool)]
+struct Pause;
+
+#[derive(Component)]
+struct PauseIndicator;
```
Shatur: For clarity. This will make it immediately clear what it is for:
```suggestion
struct PauseText;
```

examples/local_multiplayer.rs
```diff
@@ -106,6 +109,32 @@ fn setup(
...
+        actions!(
+            Shared[(
+                Action::<Pause>::new(),
+                Press::new(1.0),
+                bindings![KeyCode::Escape, GamepadButton::Start]
+            )]
+        ),
+    ));
+
+    // Pause indicator
```
Shatur: I'd remove this comment since it doesn't add anything.
```suggestion
```

examples/local_multiplayer.rs
```diff
@@ -106,6 +109,32 @@ fn setup(
...
+    commands.spawn((
+        PauseIndicator,
+        Node {
+            width: percent(100),
+            height: percent(100),
+            align_items: AlignItems::Center,
+            justify_content: JustifyContent::Center,
+            display: Display::None,
+            ..default()
```
Shatur: For consistency with the rest of the codebase. I prefer `Default` since it what RA suggests when I type `..`. In the future the `default` fn most likely will be deprecated. It was introduced to cut on the boilerplate since nesting was the main ECS API in Bevy. But with BSN introduction it won't be needed anymore.
```suggestion
            ..Default::default()
```

examples/local_multiplayer.rs
```diff
@@ -256,6 +285,21 @@ fn advance_physics(time: Res<Time>, players: Query<(&mut Transform, &mut PlayerP
     }
 }
 
+fn toggle_pause(
+    _: On<Fire<Pause>>,
+    mut time: ResMut<Time<Virtual>>,
+    mut indicator: Single<&mut Node, With<PauseIndicator>>,
+) {
+    let paused = time.is_paused();
+    if paused {
```
Shatur:
```suggestion
    if time.is_paused() {
```

examples/local_multiplayer.rs
```diff
@@ -167,7 +196,7 @@ fn player_bundle(
             context.spawn((
                 Action::<Roll>::new(),
                 DeadZone::default(),
-                DeltaScale::default(),
+                DeltaScale::virtual_time(),
```
Shatur: I'd keep the default.

examples/local_multiplayer.rs
```diff
@@ -285,6 +329,16 @@ struct AccumulatedInput {
     kick: Option<Vec2>,
 }
 
+#[derive(Component)]
+struct Shared;
```
Shatur: How about this name?
```suggestion
#[derive(Component)]
struct SharedControls;
```

Shatur (changes requested):
Thank you a lot for looking into this. I made a few comments mostly around styling and one important about explicit `Virtual`. I think it should be just `Time` to make it work correctly with the fixed schedule.

examples/character_controller.rs
jayhwkns: My bad, I thought it was my comment for some reason. I did feel like it was misleading though because typically a kinematic character controller would do something like

```
velocity += acceleration * delta_time
// Later
position += velocity * delta_time
```

Whereas this example is more like

```
// input is smoothed to simulate acceleration and friction instead of using delta time
velocity = smoothed_input
// Later
position += velocity * delta_time
```

and that's why I replaced it with the above comment to try to make it clearer. I'll put it back how it was.

src/context/time.rs
jayhwkns: I did it because it was breaking the `virtual_scaling()` test, but since that's not needed anymore I'll put it back.

The naming is a little confusing, but I'll leave it up to you whether we wanted to rename it and introduce another breaking change.

jayhwkns:
Thanks for your patience, I'll try to be more wary about the quality of what I commit. Hopefully this saves some users from some painful debugging at least.

src/context/time.rs
Shatur: Yeah, the rename is worth considering :thinking:

src/modifier/delta_scale.rs
```diff
@@ -5,17 +5,35 @@ use crate::prelude::*;
...
 }
 
+impl DeltaScale {
+    pub const REAL: Self = Self {
+        time_kind: TimeKind::Real,
+    };
+
+    pub const VIRTUAL: Self = Self {
+        time_kind: TimeKind::Virtual,
+    };
```
Shatur, line 30: Let's add docs to create links. This will make it look nicer. Here is how it's displayed right now:

Image

```suggestion
    /// Instance with [`TimeKind::Real`].
    pub const REAL: Self = Self {
        time_kind: TimeKind::Real,
    };

    /// Instance with [`TimeKind::Virtual`].
    pub const VIRTUAL: Self = Self {
        time_kind: TimeKind::Virtual,
    };
```

src/modifier/delta_scale.rs
```diff
@@ -5,17 +5,35 @@ use crate::prelude::*;
...
 #[cfg_attr(
     feature = "reflect",
     derive(Reflect),
     reflect(Clone, Component, Debug, Default)
 )]
 pub struct DeltaScale {
-    /// The type of time used to advance the timer.
+    /// The type of time used to scale the input by.
+    ///
+    /// `TimeKind::Virtual` by default for `DeltaScale` only.
```
Shatur: Let's write it this way, for consistency with other places.
```suggestion
    /// By default set to [`TimeKind::Virtual`], unlike other modifiers.
```

Shatur (changes requested):
All good, you're helping a lot 🙂 

Last a few changes, just minor suggestions for docs.

Shatur (approved):

alice-i-cecile (approved):

PR #335, opened 2026-05-12
Add support for custom binding sources

### LINKED ISSUE #31: Support for custom inputs
Right now input enum is hardcoded. We need to find a way to keep the serialization simple, but allow users extend the input map.

natepiano:
This PR adds support for custom input bindings. I use BEI for my projects and now I'm building a camera plugin that needs pinch, rotation, line/pixel wheel, and multi-touch - not yet supported by BEI. I'll probably still send small PRs adding those as proper variants later - but this PR lets me get the plugin off the ground without waiting, and it gives anyone else with a less obvious input source (VR controller, MIDI knob, network input) a way to wire it up without coming back to bei every time.

`Binding::Custom` is the new variant. `BindingSource` is the new trait. Registration mirrors `add_input_condition` and `add_input_modifier` - `binding/source.rs` is mostly a rename of `condition/fns.rs`. Update loop branches once for custom. No existing behavior changes.

I also added a `PinchGesture` example to demonstrate.

If you'd rather a different approach, just let me know and I'd be happy to change it. I copied the existing pattern because it seemed the least surprising/simplest option.


Shatur:
I definitely like the simplicity! But have you considered using input mocking for it? I just want to make sure that we don't introduce an extra concept.

natepiano:
I totally get not wanting to add conceptual surface area.  My concern with mocking (based on my read of the code, please correct me if I'm wrong)

1. it doesn't participate in apply_modifiers/apply_conditions - in the example i created i modify with Scale::splat(20.0) which is necessary for a PinchGesture.  My systems would have to read and apply the modifier myself. Also they'd have to do ActionValue, TriggerState, ActionEvents, ActionTime by hand. 

2. no way to mix custom input with built-in input - if, for example, i want to bind both the mouse_wheel and the PinchGesture to (for example) a zoom operation.  At least not flowing through the same pipeline at the same time.

I could definitely just drop this PR and make PRs for the actual input sources I'd love to see in BEI (such as PinchGesture). I just thought this Custom variant might be handy so that when someone wants to merge in a new input source, they could do it directly.  Especially if the source is really esoteric - maybe something you wouldn't want to have in the lib. 

So I'm happy either way - either you letting  this through, or you preferring targeted input source additions for things I need that would be considered more universal.

Thanks for your consideration

Shatur:
Ah, I think this makes sense. But I'd prefer a bit different approach:
- Make `Custom` hold `&'static str`.
- Create a resource that maps `&'static str` to an `ActionValue`. We probably can just expose insertion, so mappings will be created lazily.
- User fills this struct with data in any system.
- Reader reads this resource.

So everything happens through the reader. This way it will work exactly as non-builtin sources, so things like value consumption would work.

Shatur:
Bevy RC is dropped. I planned to draft a new release with the accumulated changes for 0.18 and then do the 0.19 update. But I'll probably wait for your PR first to include this into the final update for 0.18.

natepiano:
> * Make `Custom` hold `&'static str`.
> * Create a resource that maps `&'static str` to an `ActionValue`. We probably can just expose insertion, so mappings will be created lazily.
> * User fills this struct with data in any system.
> * Reader reads this resource.
> 
> So everything happens through the reader. This way it will work exactly as non-builtin sources, so things like value consumption would work.

This is much simpler! I have an update coming shortly.

natepiano:
okay i just pushed the change - please take a look and let me know if this matches what you suggested

src/context/input_reader.rs
```diff
@@ -210,6 +211,14 @@ impl InputReader<'_, '_> {
 
                 false.into()
             }
+            Binding::Custom(name) => {
+                if self.ignored(binding) {
+                    return ActionValue::Bool(false);
+                }
+                self.custom_inputs
+                    .get(name)
+                    .unwrap_or(ActionValue::Bool(false))
```
Shatur: The returned dimention should match what was returned previously. So we need to get the value first and check for ignored after that. If the value is not present, returning bool is fine.

src/context/input_reader.rs
```diff
@@ -374,6 +384,7 @@ pub(crate) struct IgnoredInputs {
     gamepad_buttons: HashSet<GamepadInput<GamepadButton>>,
     gamepad_axes: HashSet<GamepadInput<GamepadAxis>>,
     any_key: bool,
+    customs: HashSet<&'static str>,
```
Shatur: For consistency:
```suggestion
    custom_inputs: HashSet<&'static str>,
```

src/context.rs
```diff
@@ -560,8 +563,6 @@ fn update<S: ScheduleLabel>(
                         }
                     }
 
-                    let mut binding_entity = conds_and_mods.get_mut(binding_entity).unwrap();
-
```
Shatur: Could you undo these changes? Looks like leftover from the old approach.

examples/custom_binding_source.rs
```diff
@@ -0,0 +1,75 @@
+//! Maps trackpad [`PinchGesture`] to a [`Zoom`] action via [`Binding::Custom`]
```
Shatur: Let's remove the runnable example and only keep the doc one. I use examples only for things that are necessary to be runnable.

tests/custom_input.rs
```diff
@@ -0,0 +1,85 @@
+use bevy::prelude::*;
```
Shatur, line 1: Let's rename the module into `custom_input`.

tests/binding_source.rs
```diff
@@ -0,0 +1,85 @@
...
+    let mut q = app
+        .world_mut()
+        .query::<(&Action<TestValue>, &TriggerState)>();
+    let (&value, &state) = q.single(app.world()).unwrap();
+    assert_eq!(*value, 0.75);
+    assert_eq!(state, TriggerState::Fired);
+}
+
+#[test]
+fn binding_entity_modifier_applies_to_custom_input() {
```
Shatur: I think we don't need to test for it. It's part of the reader, so naturally all modifiers and conditions will work.

src/binding.rs
```diff
@@ -175,6 +179,7 @@ impl Display for Binding {
             Binding::GamepadButton(gamepad_button) => write!(f, "{gamepad_button:?}"),
             Binding::GamepadAxis(gamepad_axis) => write!(f, "{gamepad_axis:?}"),
             Binding::AnyKey => write!(f, "Any Key"),
+            Binding::Custom(name) => write!(f, "Custom({name})"),
```
Shatur: Let's just display the name, it will look nicer.
```suggestion
            Binding::Custom(name) => write!(f, "{name}"),
```

src/binding/source.rs
```diff
@@ -0,0 +1,50 @@
...
+/// fn stage_pinch(
+///     mut events: MessageReader<PinchGesture>,
+///     mut customs: ResMut<CustomInputs>,
+/// ) {
+///     let delta: f32 = events.read().map(|e| e.0).sum();
+///     customs.insert("pinch", ActionValue::Axis1D(delta));
+/// }
+/// ```
+#[derive(Resource, Default, Debug)]
+pub struct CustomInputs(HashMap<&'static str, ActionValue>);
```
Shatur: Maybe derive `Deref` and `DerefMut`? This way we probably won't need other methods and a dedicated module.

tests/binding_source.rs
```diff
@@ -0,0 +1,85 @@
...
+
+    app.update();
+
+    let mut q = app.world_mut().query::<&Action<TestValue>>();
+    let &value = q.single(app.world()).unwrap();
+    assert_eq!(*value, -0.5);
+}
+
+#[test]
+fn unset_custom_input_reads_zero() {
```
Shatur: I prefer shorter test names. For example, here we're inside `custom` module (after the rename), so it's obvious that all tests will be for custom input. And I usually don't encode the behavior, just try to make the test obvious.

So I'd call it like this:

```suggestion
fn no_value() {
```

tests/binding_source.rs
```diff
@@ -0,0 +1,85 @@
...
+        .add_input_context::<TestContext>()
+        .finish();
+
+    app.world_mut()
+        .resource_mut::<CustomInputs>()
+        .insert(TEST, ActionValue::Axis1D(0.75));
+
+    app.world_mut().spawn((
+        TestContext,
+        actions!(TestContext[(Action::<TestValue>::new(), bindings![Binding::Custom(TEST)],)]),
```
Shatur:
```suggestion
        actions!(TestContext[(Action::<TestValue>::new(), bindings![Binding::Custom(TEST)])]),
```

tests/binding_source.rs
```diff
@@ -0,0 +1,85 @@
+use bevy::prelude::*;
+use bevy_enhanced_input::prelude::*;
+use test_log::test;
+
+const TEST: &str = "test";
+
+#[test]
+fn custom_input_value_reaches_action() {
```
Shatur: In contrast to `no_value`. Maybe worth swapping them places.
```suggestion
fn with_value() {
```

src/binding.rs
```diff
@@ -105,6 +106,8 @@ pub enum Binding {
     /// inputs, not just the one that activated it. To have an action with this binding
     /// evaluated first, place it in a higher-priority context.
     AnyKey,
+    /// Value read from the [`CustomInputs`](source::CustomInputs) resource by name.
+    Custom(&'static str),
```
Shatur:
```suggestion
    /// Name of the input entry in [`CustomInputs`](source::CustomInputs) to read.
    ///
    /// Use this for inputs that can't be represented or feed into Bevy input resources.
    Custom(&'static str),
```

src/binding/source.rs
```diff
@@ -0,0 +1,50 @@
+//! Values for [`Binding::Custom`] entries.
+
+use bevy::{platform::collections::HashMap, prelude::*};
+
+use crate::prelude::*;
+
+/// Stores values for [`Binding::Custom`] entries, keyed by name.
+///
+/// Write to this resource from any system. A missing entry reads as
+/// [`ActionValue::Bool`] `false`.
```
Shatur:
```suggestion
/// Write to this resource from any system to make custom input values available to actions.
///
/// Missing entries are read as [`ActionValue::Bool`] `false`.
```

src/binding/source.rs
```diff
@@ -0,0 +1,50 @@
+//! Values for [`Binding::Custom`] entries.
+
+use bevy::{platform::collections::HashMap, prelude::*};
+
+use crate::prelude::*;
+
+/// Stores values for [`Binding::Custom`] entries, keyed by name.
```
Shatur:
```suggestion
/// Stores values for [`Binding::Custom`] entries.
```

src/binding/source.rs
```diff
@@ -0,0 +1,50 @@
...
+/// use bevy::{input::gestures::PinchGesture, prelude::*};
+/// use bevy_enhanced_input::prelude::*;
+///
+/// fn stage_pinch(
+///     mut events: MessageReader<PinchGesture>,
+///     mut customs: ResMut<CustomInputs>,
+/// ) {
+///     let delta: f32 = events.read().map(|e| e.0).sum();
+///     customs.insert("pinch", ActionValue::Axis1D(delta));
+/// }
```
Shatur: Maybe worth showing the system registration? I.e. it should be after Bevy's input system, but before BEI.

Shatur (changes requested):
Thanks! Left suggestions, mostly nits.

src/binding/source.rs
```diff
@@ -0,0 +1,33 @@
+//! Stores values for [`Binding::Custom`] entries.
+
+use bevy::{platform::collections::HashMap, prelude::*};
+
+use crate::prelude::*;
+
+/// Write to this resource from any system to make custom input values available to actions.
```
Shatur: It's now a module with a single struct, maybe put into `reader`?

Shatur (changes requested):
One last suggestion 🙂 And CI, of course.

Shatur (approved):

natepiano:
feature combinations broke - i could use `Binding::Custom(Cow<'static, str>)`  and wire it through the changes 
or
#[serde(skip)] the Custom variant
any preference?

I'll also fix the rustdoc issue

Shatur:
Ah, I totally forgot about this! The problem with `Cow` is that we lose `Copy` on `Binding`, which is super annoying.
And serialization of this field also important.

How about this:
- Create `CustomInput` which holds `usize`.
- Store it inside our `Custom`.
- Inside `CustomInput` store a counter that starts from 0 and a function called `register_input` which increments the counter and returns `CustomInput`.

We'll have to change `Display` for this variant to `{input:?}`, but serialization will continue to work and since it's easy to make registrations deterministic, it should work.

And I'll have to ask to move `CustomInput` back into its own module (let's call it `custom`) under `reader`.

Thoughts?

natepiano:
no problem - i'll take a crack at it - think i can push it tonight (eastern time) tomorrow for sure

Shatur:
No problem, I'll wait :slightly_smiling_face:

natepiano:
just pushed

natepiano:
pushed again to use `bevy::platform AtomicUsize` - `core::sync::atomic` was failing on `thumbv6m-none-eabi` target - trying again

src/context/input_reader/custom.rs
```diff
@@ -0,0 +1,80 @@
...
+impl CustomInput {
+    /// Returns a fresh identifier.
+    ///
+    /// Each call increments a process-wide counter. IDs are assigned in
+    /// registration order, so two runs that register in the same order
+    /// produce the same IDs.
+    #[must_use]
+    pub fn register_input() -> Self {
+        static COUNTER: AtomicUsize = AtomicUsize::new(0);
+        Self(COUNTER.fetch_add(1, Ordering::Relaxed))
```
Shatur: No, no, we don't need any atomics here. I were talking about a simple counter field inside `CustomInputs`. Just switch from tuple to a named struct with `map: HashMap` and `counter: usize`.

src/context/input_reader/custom.rs
```diff
@@ -0,0 +1,80 @@
...
+    feature = "reflect",
+    derive(Reflect),
+    reflect(Clone, Debug, Hash, PartialEq)
+)]
+#[cfg_attr(feature = "serialize", derive(Serialize, Deserialize))]
+#[cfg_attr(
+    all(feature = "reflect", feature = "serialize"),
+    reflect(Serialize, Deserialize)
+)]
+#[cfg_attr(feature = "serialize", serde(transparent))]
```
Shatur: Unnecessary:
```suggestion
```

src/context/input_reader/custom.rs
```diff
@@ -0,0 +1,80 @@
...
+};
+#[cfg(feature = "serialize")]
+use serde::{Deserialize, Serialize};
+
+use crate::prelude::*;
+
+/// Identifier for a custom input, used as the payload of [`Binding::Custom`].
+///
+/// Obtain one with [`CustomInput::register_input`] at startup and reuse it
+/// when binding and when writing values into [`CustomInputs`].
```
Shatur: Let's keep it short and just link to the `CustomInputs` with full documentation:
```suggestion
/// Identifier for a custom input, used in [`Binding::Custom`].
///
/// Obtainable from [`CustomInputs::register_input`].
```

src/context/input_reader/custom.rs
```diff
@@ -0,0 +1,80 @@
...
+    /// produce the same IDs.
+    #[must_use]
+    pub fn register_input() -> Self {
+        static COUNTER: AtomicUsize = AtomicUsize::new(0);
+        Self(COUNTER.fetch_add(1, Ordering::Relaxed))
+    }
+}
+
+/// Write to this resource from any system to make custom input values available to actions.
+///
```
Shatur, line 28: Let's add something like this:
```suggestion
///
/// To register a custom input, use [`Self::register_input`].
```

Based on my suggestion about the counter.

src/context/input_reader/custom.rs
```diff
@@ -0,0 +1,80 @@
...
+/// ```
+/// use bevy::{input::gestures::PinchGesture, prelude::*};
+/// use bevy_enhanced_input::prelude::*;
+///
+/// #[derive(Resource)]
+/// struct PinchId(CustomInput);
+///
+/// let mut app = App::new();
+/// let pinch = CustomInput::register_input();
```
Shatur: After my suggestion you'll get it from the resource. Don't forget to add plugins to the app, otherwise the doc test will panic.

src/context/input_reader/custom.rs
```diff
@@ -0,0 +1,80 @@
+//! Identifiers and storage for [`Binding::Custom`] entries.
```
Shatur:
```suggestion
//! Identifiers and storage for [`Binding::Custom`] values.
```

Shatur (changes requested):
Left a few more suggestions, but I think we're very close.

natepiano:
okay - this is simpler again - thanks for iterating on this, I appreciate it

src/context/input_reader/custom.rs
```diff
@@ -0,0 +1,84 @@
...
+)]
+pub struct CustomInput(usize);
+
+/// Stores values for [`Binding::Custom`] entries.
+///
+/// Write to this resource from any system to make custom input values available to actions.
+///
+/// Missing entries are read as [`ActionValue::Bool`] `false`.
+///
+/// To register a custom input, use [`Self::register_input`].
```
Shatur, line 29: Could you put this line right after `Write to this resource`? I think it will read better.

src/context/input_reader/custom.rs
```diff
@@ -0,0 +1,84 @@
...
+///
+/// Feeding trackpad pinch events into a `Binding::Custom` entry:
+///
+/// ```
+/// use bevy::{input::gestures::PinchGesture, prelude::*};
+/// use bevy_enhanced_input::prelude::*;
+///
+/// #[derive(Resource)]
+/// struct PinchId(CustomInput);
+///
```
Shatur: I use top to bottom ordering in this crate. I.e. high level things come first. So declarations is something that I put at the end. Could you move this to the end of the example?

src/context/input_reader/custom.rs
```diff
@@ -0,0 +1,84 @@
...
+/// Write to this resource from any system to make custom input values available to actions.
+///
+/// Missing entries are read as [`ActionValue::Bool`] `false`.
+///
+/// To register a custom input, use [`Self::register_input`].
+///
+/// # Examples
+///
+/// Feeding trackpad pinch events into a `Binding::Custom` entry:
```
Shatur:
```suggestion
/// Feeding trackpad pinch events:
```

src/context/input_reader/custom.rs
```diff
@@ -0,0 +1,84 @@
...
+    #[deref]
+    map: HashMap<CustomInput, ActionValue>,
+    counter: usize,
+}
+
+impl CustomInputs {
+    /// Returns a fresh identifier.
+    ///
+    /// IDs are assigned in registration order, so two runs that register in
+    /// the same order produce the same IDs.
```
Shatur:
```suggestion
    /// Creates a new custom input identifier.
    ///
    /// IDs are assigned in registration order, so two runs that register in
    /// the same order produce the same IDs.
```

Shatur (changes requested):
Final pass and we're good to go 🙂

natepiano:
updated

Shatur (approved):

alice-i-cecile (approved):

PR #336, opened 2026-05-17
Fixed misleading notes in CHANGELOG.md

jayhwkns:
We decided on the constants approach instead of a constructor approach mid-PR. My bad for not updating the changelog alongside that decision.


Shatur (approved):

Shatur:
Ah, thanks!

PR #339, opened 2026-06-08
Fix input accumulation in local_multiplayer example

jayhwkns:
Turns out I had a slight misinterpretation of what it meant to "accumulate input" when I first wrote this example 😓

This misconception gave me some trouble in my own project, so I wanted to ensure it was correct here. Don't ask why I got it correct in character controller, but not here. I wish I could say.


Shatur (approved):
Ah, of course, it's the direction! Thanks for the fix!

PR #346, opened 2026-07-24
Upgrade keybinding_menu example to use Bevy 0.19's settings

mgi388:
**Test plan**

- Ran the example: `cargo run --example keybinding_menu --features="serialize"`
- Looked at the output of the settings file (it doesn't exist yet):

```
❯ cat {me}/Library/Preferences/com.github.simgine.bevy_enhanced_input.examples.keybinding_menu/settings.toml
cat: {me}/Library/Preferences/com.github.simgine.bevy_enhanced_input.examples.keybinding_menu/settings.toml: No such file or directory
```

- Pressed **Apply** in the UI and looked at the file again (it now exists):

```ron
[input_settings]
backward = [{ Keyboard = { key = "KeyS", mod_keys = "" } }, { Keyboard = { key = "ArrowDown", mod_keys = "" } }, "None"]
fire = ["None", "None", "None"]
forward = [{ Keyboard = { key = "KeyW", mod_keys = "" } }, { Keyboard = { key = "ArrowUp", mod_keys = "" } }, "None"]
jump = [{ Keyboard = { key = "Space", mod_keys = "" } }, "None", "None"]
left = [{ Keyboard = { key = "KeyA", mod_keys = "" } }, { Keyboard = { key = "ArrowLeft", mod_keys = "" } }, "None"]
right = [{ Keyboard = { key = "KeyD", mod_keys = "" } }, { Keyboard = { key = "ArrowRight", mod_keys = "" } }, "None"]
run = [{ Keyboard = { key = "ShiftLeft", mod_keys = "" } }, "None", "None"]
```

- Changed one of the bindings in the UI (**Run** rebound to LMB), checked the file was still the same (I didn't apply yet), then applied and checked the file was now updated:

```ron
[input_settings]
backward = [{ Keyboard = { key = "KeyS", mod_keys = "" } }, { Keyboard = { key = "ArrowDown", mod_keys = "" } }, "None"]
fire = ["None", "None", "None"]
forward = [{ Keyboard = { key = "KeyW", mod_keys = "" } }, { Keyboard = { key = "ArrowUp", mod_keys = "" } }, "None"]
jump = [{ Keyboard = { key = "Space", mod_keys = "" } }, "None", "None"]
left = [{ Keyboard = { key = "KeyA", mod_keys = "" } }, { Keyboard = { key = "ArrowLeft", mod_keys = "" } }, "None"]
right = [{ Keyboard = { key = "KeyD", mod_keys = "" } }, { Keyboard = { key = "ArrowRight", mod_keys = "" } }, "None"]
run = [{ MouseButton = { button = "Left", mod_keys = "" } }, "None", "None"]
```

**Notes**

- Simplified the `dev-dependencies` Bevy features because settings wouldn't work without adding some extra ones (reflect ones I think), and it was simpler to just use Bevy's new collections, plus add in `bevy_settings`.
- The reverse domain is verbose, but it's accurate AFAIK. When BEI is upstreamed, it will become similar to `org.bevy.examples.settings` (the one the Bevy `settings` example uses).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,22 +41,11 @@ serde = { version = "1.0", default-features = false, features = [
 bitflags = { version = "2.6", default-features = false, features = ["serde"] }
 
 [dev-dependencies]
-bevy = { version = "0.19.0", default-features = false, features = [
-  "bevy_gilrs",
-  "bevy_log",
-  "bevy_pbr",
-  "bevy_state",
-  "bevy_ui_render",
-  "bevy_window",
-  "default_font",
-  "debug",
-  "tonemapping_luts",
-  "ui_picking",
-  "x11",
-  "zstd_rust",
+bevy = { version = "0.19.0", features = [
+  "2d", "3d", "ui",
+  "bevy_settings",
 ] }
 test-log = "0.2"
-ron = "0.12"
 
 [lints.clippy]
 type_complexity = "allow"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,10 +41,7 @@ serde = { version = "1.0", default-features = false, features = [
 bitflags = { version = "2.6", default-features = false, features = ["serde"] }
 
 [dev-dependencies]
-bevy = { version = "0.19.0", features = [
-  "2d", "3d", "ui",
-  "bevy_settings",
-] }
+bevy = { version = "0.19.0", features = ["2d", "3d", "ui", "bevy_settings"] }
 test-log = "0.2"
 
 [lints.clippy]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,22 @@ serde = { version = "1.0", default-features = false, features = [
 bitflags = { version = "2.6", default-features = false, features = ["serde"] }
 
 [dev-dependencies]
-bevy = { version = "0.19.0", features = ["2d", "3d", "ui", "bevy_settings"] }
+bevy = { version = "0.19.0", default-features = false, features = [
+  "bevy_gilrs",
+  "bevy_log",
+  "bevy_pbr",
+  "bevy_settings",
+  "bevy_state",
+  "bevy_ui_render",
+  "bevy_window",
+  "default_font",
+  "debug",
+  "reflect_auto_register", # TODO: Remove this once `bevy_support` correctly depends on it
+  "tonemapping_luts",
+  "ui_picking",
+  "x11",
+  "zstd_rust",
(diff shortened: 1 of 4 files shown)


alice-i-cecile (approved):
Nice!

Cargo.toml
```diff
@@ -41,22 +41,8 @@ serde = { version = "1.0", default-features = false, features = [
...
-  "bevy_state",
-  "bevy_ui_render",
-  "bevy_window",
-  "default_font",
-  "debug",
-  "tonemapping_luts",
-  "ui_picking",
-  "x11",
-  "zstd_rust",
-] }
```
Shatur, line 57: We should depend only on things that actually used by examples.

Shatur (changes requested):

Cargo.toml
mgi388: I agree for non-dev dependencies, but I tend to disagree for examples. Compare the before and after diff in this PR and maintenance is just simpler when you use the new Bevy feature collections for examples. 

I think when upstreamed to Bevy proper, this doesn't matter either.

I'm also just not keen on hunting down the missing feature(s) that was making `bevy_settings` fail when we could just use collections instead but I know that's lazy :) 

Up to you though.

Shatur: It takes much longer to compile and requires installing additional dependencies on CI. Until the crate is upstreamed, it's just easier for me to maintain the crate this way.

If enabling `bevy_settings` on `bevy` doesn't work properly, it's a bug in Bevy.

examples/keybinding_menu.rs
```diff
@@ -381,10 +372,7 @@ fn apply(
 
     commands.trigger(SettingsChanged);
 
-    match settings.write(SETTINGS_PATH) {
-        Ok(()) => info!("writing settings to '{SETTINGS_PATH}'"),
-        Err(e) => error!("unable to write settings to '{SETTINGS_PATH}': {e}"),
-    }
+    commands.queue(SaveSettingsDeferred(Duration::from_secs_f32(0.1)));
```
Shatur: Why deferred instead of sync? The description says `SaveSettingsDeferred` is for settings that change frequently.

Shatur (changes requested):
One last question 🙂

examples/keybinding_menu.rs
mgi388: Good pickup, that was my mistake.

Shatur (approved):

